PR #4, opened 2015-04-16
Add Vagrantfile and provison.sh script

chrismytton:
I've also added a missing `rack-test` dependency to the `Gemfile` which was preventing the tests from passing.


scripts/provision.sh
```diff
@@ -0,0 +1,36 @@
...
+# Install application gems
+sudo gem install bundler --no-rdoc --no-ri
+bundle install
+
+# Set shell login message
+echo "-------------------------------------------------------
+Welcome to the popolo-viewer-sinatra vagrant machine
+
+Run the web server with:
+  bundle exec rackup -o 0.0.0.0
```
tmtmtmtm: Probably worth adding a note here that you'll need to restart it for app changes (though not template changes)

tmtmtmtm:
:+1:

PR #6, opened 2015-04-16
Example of using scss to render stylesheets

chrismytton:
This adds the sass gem then uses sinatra's build in scss handling to render the stylesheet at the given route. Ideally in a production environment this should actually be a static file being served from the filesystem, or should at least have some caching headers. But for development it should be sufficient.

cc @zarino


chrismytton:
Closing as these commits are now part of the [ssues/5-everypolitician-styling](https://github.com/mysociety/popolo-viewer-sinatra/compare/issues/5-everypolitician-styling) branch.

PR #12, opened 2015-04-17
Style guide

### LINKED ISSUE #8: Styling crib-sheet for future contributors
As of https://github.com/mysociety/popolo-viewer-sinatra/pull/7, the site uses an EveryPolitician-like stylesheet, with no underlying third-party framework.

This is awesome, because it's ultra lightweight (only basic typography and the new stuff we define are covered). But it's also difficult, because non-designers can no longer just copy-and-paste examples from, say, the Bootstrap docs.

Rather than refactoring the styles to be built on top of Bootstrap, let's just document what we've got, so people know what they can use.

Eg: Styles we've got that won't be immediately obvious to contributors:

* `p.lead` – larger paragraph style, ideal for use under headings
* `.button` and modifier classes like `.button--secondary`
* `.dual-buttons`, `.light-page-section` and other common components

Then there's also guidelines on how to make a column layout, or how to make a hero section, or how to mark up a code example.

zarino:
Fixes #8. 

I've refactored lots of the Sass, added a "Style guide" for non-designers, and added a Readme telling people about Vagrant, Sass, and the new style guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,22 @@
+# Install
+
+Local development requires [Vagrant](http://docs.vagrantup.com/v2/installation/).
+
+    git clone https://github.com/mysociety/popolo-viewer-sinatra.git
+    cd popolo-viewer-sinatra
+    vagrant up
+    vagrant ssh
+
+Then follow the instructions displayed by the virtual machine.
+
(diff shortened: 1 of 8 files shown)


tmtmtmtm:
@zarino This style guide is superb!!

PR #36, opened 2015-04-21
Style the list of people on a term page.

### LINKED ISSUE #33: Switch /term/X page to the new layout
Still just has an unstyled list of people:

term page 2015-04-21 at 13 21 38

This should have the same styling as on /finland/people.html, for example

duncanparkes:
Give the list of people on a term page the same styling
as the list on the people page.

Fixes #33.

### DIFF
--- a/views/term.erb
+++ b/views/term.erb
@@ -10,8 +10,8 @@
                <a href="<%= person_url(m['person']) %>">
                  <span class="avatar"><i class="fa fa-user"></i></span>
                  <span class="label"><%= m['person']['name'] %></span>
-                 (<%= m['start_date'] %>–<%= m['end_date'] %>)
                </a>
+               (<%= m['start_date'] %>–<%= m['end_date'] %>)
             </li>
           <% end %>
         </ul>
--- a/views/term.erb
+++ b/views/term.erb
@@ -4,11 +4,14 @@
         <p><%= @term['start_date'] %> – <%= @term['end_date'] %></p>
 
       <% if @memberships %>
-        <ul>
+        <ul class="avatar-list">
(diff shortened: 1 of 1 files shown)


views/term.erb
```diff
@@ -4,11 +4,14 @@
...
+        <ul class="avatar-list">
           <% @memberships.sort_by { |m| m['person']['name'] }.each do |m| %>
             <li>
-                <a href="<%= person_url(m['person']) %>"><%= m['person']['name'] %></a>
-                (<%= m['start_date'] %>–<%= m['end_date'] %>)
+               <a href="<%= person_url(m['person']) %>">
+                 <span class="avatar"><i class="fa fa-user"></i></span>
+                 <span class="label"><%= m['person']['name'] %></span>
+                 (<%= m['start_date'] %>–<%= m['end_date'] %>)
+               </a>
```
tmtmtmtm: the link should really close _before_ the dates, rather than after, partially because it makes more sense to (as the dates might want to be links to something else), but also because it causes some styling confusion otherwise (the names get an underscore on hover, but the dates don't)

duncanparkes: Good point!

duncanparkes:
Moving the date outside the link causes everything to line break between the name and the date for me.

tmtmtmtm:
Yeah, that's pretty nasty. Hopefully @zarino knows the magical incantation ...

zarino:
@duncanparkes Yes this is a bit tricky. It'll require me to beef up the `.avatar-list` component to allow for a line of content below the person's name.

@tmtmtmtm Could you create a new ticket for that and assign to me?

tmtmtmtm:
@zarino: That's opened as #38, thanks.

I'll merge this for now on the assumption that everything will Just Work after that :)

PR #44, opened 2015-04-22
Only show dates for memberships if they don't match the term.

### LINKED ISSUE #34: On /term/X page, only show dates if different to that of the Term
There's no need to show the dates for everyone if the majority will be the same as the Term itself

term page 2015-04-21 at 13 21 38

(Probably also want to check that those dates are within the term)

duncanparkes:
On the term page, only show dates for a membership if those differ from the dates of the term.

Fixes #34


tmtmtmtm:
:+1:

PR #54, opened 2015-04-23
Sinatra routes

tmtmtmtm:
Sinatra has built-in magic `__sinatra__` routes, for things like
images. We need to allow those to pass through our before filter.


tmtmtmtm:
Closing in favour of #55

PR #57, opened 2015-04-23
Only show legislative memberships on a Person page

### LINKED ISSUE #51: /wales/person/169 Errors
screen shot 2015-04-23 at 08 59 22

tmtmtmtm:
A Person page is breaking when viewing someone who’s in both the Legislature and the Executive

As we’re not really handling Executive positions at all yet, make this display only legislative memberships for now.

Before:
screen shot 2015-04-23 at 08 59 22

After:
david melding 2015-04-23 at 09 58 01

Fixes #51

### DIFF
--- a/t/web/wales.rb
+++ b/t/web/wales.rb
@@ -1,0 +1,37 @@
+ENV['RACK_ENV'] = 'test'
+
+require_relative '../../app'
+require 'minitest/autorun'
+require 'rack/test'
+require 'nokogiri'
+
+include Rack::Test::Methods
+
+def app
+  Sinatra::Application
+end
+
+describe "Wales" do
+
+  subject { Nokogiri::HTML(last_response.body) }
+
+  #-------------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


t/web/wales.rb
```diff
@@ -0,0 +1,37 @@
...
+    before { get '/wales/person/169' }
+
+    it "should have have their name" do
+      subject.css('h1').text.must_equal 'David Melding'
+    end
+
+    it "should have have their legislative membership" do
+      subject.css('ul li').inner_html.must_include 'Welsh Conservative Party'
+    end
+
```
duncanparkes: Is it worth checking the lack of executive membership here? Or are you happy just doing that in the helper?

duncanparkes:
:+1: Looks good to me!

t/web/wales.rb
tmtmtmtm: I think I'm OK just leaving that to the helper until we actually add executive memberships. This test is mostly just to check that the page loads at all and displays the information we _do_ expect, as it was 500-ing before this fix.

PR #62, opened 2015-04-23
Development Mode test

### LINKED ISSUE #56: Run a single test in Development mode
In #55 I wanted to add a test to `t/web/basic.rb` like:

```
  describe "magic sinatra routes" do

    before { get '/__sinatra__/500.png' }

    it "should pass __sinatra__ requests through" do
      last_response.status.must_equal 200
    end
  end
```

But this fails, because the test explicitly sets `ENV['RACK_ENV'] = 'test'`, and this will only work in `development`. I couldn't find a way to override that just for a single test, and I'm not sure what the implications are of setting it for the whole file.

tmtmtmtm:
In #55 I made it so that we don’t clobber the magic `/__sinatra__/` routes, but couldn't work out how to add a test for that.

After talking about it with @chrismytton, it seems that the best way is to just have a separate `development` mode test, as the routes are getting created long before the test runs.

Closes #56

### DIFF
--- a/t/web/basic.rb
+++ b/t/web/basic.rb
@@ -44,15 +44,5 @@ describe "Viewer" do
     end
   end
 
-  describe "magic sinatra routes" do
-
-    before { get '/__sinatra__/500.png' }
-
-    it "should pass __sinatra__ requests through" do
-      last_response.status.must_equal 200
-    end
-  end
-  
-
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
Closing in favour of #63

Issue #96, opened 2015-04-27
Term Table design


tmtmtmtm:
We currently have a simple table of members during a term, e.g. `/finland/term_table`

![term table 2015-04-27 at 14 54 51](https://cloud.githubusercontent.com/assets/57483/7348924/679993dc-eced-11e4-83ac-3fe63a38c6a2.png)

Party changes, etc, are simply shown as grouped rows:
![change party 2015-04-27 at 14 55 19](https://cloud.githubusercontent.com/assets/57483/7348938/79db2fc4-eced-11e4-8712-0c8c3bea8ea8.png)

Things we'll want to add to this include:
- [x] Handling bicameral legislatures, perhaps with a 'tabbed' interface to switch between the houses. #97 
- [x] A display for the current breakdown of Parties  — Conservative 302, Labour 256, Liberal Democrats 56, etc (where possible, Government Parties will be split out/highlighted separately) #98 
- [x] Filtering/Sorting the list of People
- [x] Where we have data for other terms, simple navigation to those, either just in forward/back buttons, or perhaps a little table like at the bottom of similar Wikipedia pages, e.g. https://en.wikipedia.org/wiki/List_of_members_of_the_Folketing#References #101 
- [x] a "Download this as CSV" button #99

tmtmtmtm:
@zarino — In case it impacts any of your thinking, this is the main page that will go live for every country *before* any of the other pages — i.e. this should to be capable of standing on its own without links to Person or Party pages etc initially, until "Stage 2" for that country is ready, as well as working when those things *do* become links.

(However, if required, we _could_ have two versions of the page, depending on whether the Country is at the next stage or not. I'd prefer not to, but if the design works better like that, then it's doable.)

zarino:
I've started design tweaks in the branch [issues/96-term-table-design](https://github.com/mysociety/popolo-viewer-sinatra/compare/issues/96-term-table-design?expand=1)

PR #122, opened 2015-04-30
Sortable table plugin

zarino:
For #96.

Add a class to a table, and it becomes sortable!

Works with [fixed headers](https://github.com/mysociety/popolo-viewer-sinatra/commit/472768fa5acb06b9c3b983d63a07ded428f87505) too.

sortable

### DIFF
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -71,6 +71,115 @@
 
 }(jQuery));
 
+(function ($) {
+
+  $.fn.sortable = function() {
+
+    // Call this on a <table> and it'll make all the columns sortable.
+
+    var sortTable = function sortTable($table, columnIndex){
+      // Find the rows to sort, and their parent
+      var $tbody = $table.children('tbody');
+      if($tbody.length == 0){ $tbody = $table; }
+      var $trs = $tbody.children('tr');
+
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
Woohoo! :rocket:

tmtmtmtm:
Erk. This breaks quite badly for Finland.

tmtmtmtm:
I'm assuming it's to do with the rowspans...

zarino:
Oh no! Rowspans!

I guess, the first time you sort the table, it's going to have to un-span the double-height cells (and duplicate their content into all the now missing cells below) so the rows can be sorted independently. Unfortunately, it means, once you return back to the original (unsorted) order, the rowspans will have gone. But at least all the data will be there. Would you be happy with that @tmtmtmtm?

tmtmtmtm:
Perhaps then that should be the default position? i.e. have the generated table always show full rows by default (no spans). Then we have a JS-function that can do the combining if the table is sorted by name, and consecutive rows are the same person (there's already a `mem-{id}` set on the row). Then we just trigger that on initial load, or when sorting by name, and detrigger it before any other sort?

(BTW: I've pushed a hotfix to disable it just for Finland on the live site)

PR #166, opened 2015-05-10
Add a 'download as JSON' button

### LINKED ISSUE #146: JSON Download
Similarly to the CSV download, we should also offer a location to download the (post-processed) Popolo JSON we're using for each country. It _might_ be valuable to offer this on a per-Term basis, but initially it's probably enough to just offer the whole file.

tmtmtmtm:
If we move the JSON into public/data, we can just pass it straight-through. Closes #146 

----
Before:
csv download 2015-05-10 at 18 45 06

After:
json download 2015-05-10 at 18 41 53


tmtmtmtm:
@zarino I find it slightly odd that the buttons split to opposite sides. I'm open to persuasion that that's what they *should* do (or that we should do something else entirely here), but otherwise, is it a simple fix to have them group together (probably at the right)?

PR #177, opened 2015-05-13
Add select box for quickly jumping between terms

zarino:
Looks like this for countries with more than one term:

screen shot 2015-05-13 at 17 58 18

screen shot 2015-05-13 at 17 58 29

Stacks vertically on narrow screens:

screen shot 2015-05-13 at 17 59 33


tmtmtmtm:
@zarino: Thanks! :+1:

PR #207, opened 2015-05-26
Use https for everything

tmtmtmtm:
Getting jQuery and Google Fonts over http means that they don't work if
we're viewing the site over https (e.g. on heroku). Switch to always
using the https: versions.


dracos:
For interest, in general, you can skip the scheme part of the URL, ie. use `//fonts.googleapis.com/...` and it will use the same scheme as the page it's currently on.

tmtmtmtm:
*nod* — we talked about doing that, but @chrismytton pointed me at http://www.paulirish.com/2010/the-protocol-relative-url/ which convinced me to just make it https everywhere.

PR #227, opened 2015-06-02
Rebuild everypolitician/viewer-static on Travis-CI

chrismytton:
When travis does a successful build of the master branch (not as part of a pull request) it rebuilds the viewer-static:gh-pages branch, which is hosted at data.everypolitician.org.

This pull request also adds bundler caching to .travis.yml, which is available on container based builds (which this repository uses).


tmtmtmtm:
I haven't used deploykeys before, so don't really understand what's going on here. Is this just copying what's documented at https://developer.github.com/guides/managing-deploy-keys/ ?

tmtmtmtm:
What happens if we merge several things in quick succession? Is there a potential race condition when the later one finishes first and then gets overwritten by the earlier one?

chrismytton:
> I haven't used deploykeys before, so don't really understand what's going on here. Is this just copying what's documented at https://developer.github.com/guides/managing-deploy-keys/ ?

Not exactly. I've basically cargo-culted the [custom deployment with git](http://docs.travis-ci.com/user/deployment/custom/#Git) example in the travis docs and added an SSH key to the @everypoliticianbot account. The SSH private key is encrypted as `deploy_key.pem.enc` and gets decrypted in the `release.sh` script so it can be used for pushing.

Using an SSH key allows us to push over SSH instead of https, which is preferable because pushing over https can (and did during testing!) lead to accidentally leaking access tokens/passwords.



> What happens if we merge several things in quick succession? Is there a potential race condition when the later one finishes first and then gets overwritten by the earlier one?

Hmm, good question! My understanding of travis is that builds are run serially, so a new one won't start until the previous one has finished, but I can't actually find a definite answer to this in the docs. The closest I've found is [Speeding up the build](http://docs.travis-ci.com/user/speeding-up-the-build/), which seems to suggest that the only way to get the builds to run in parallel is to use [The Build Matrix](http://docs.travis-ci.com/user/build-configuration/#The-Build-Matrix).

tmtmtmtm:
> I've basically cargo-culted the custom deployment with git example in the travis docs and added an SSH key to the @everypoliticianbot account  

Where's the best place to document all this? I uncovered some of it by googling for what _might_ be happening, and then working backwards, but we should probably have a better explanation that that (or assuming that someone interested will think to look at the PR for the code that they're confused by). Just add something to the README? 

[And, yes, this problem isn't unique to here — the whole way that all the repos hang together isn't entirely obvious, though hopefully splitting out https://github.com/everypolitician/viewer-static will help a little.]

> builds are run serially, so a new one won't start until the previous one has finished

Ah, that would make a lot of sense. I guess there's a potential problem in the future if they change that, but that's yagni for now, I suspect.


So :+1: on this (other than docs, which can happen separately).

tmtmtmtm:
(Will merge this after some cleanups I'm doing on the static site separately)

Issue #279, opened 2015-06-18
Redesign homepage


wrightmartin:
- Make it look more like a mySociety project
- Incorporate the brand @wrightmartin started working on

![every-p](https://cloud.githubusercontent.com/assets/2292925/8234260/c97fb992-15d2-11e5-9a7d-1466d404ca7b.png)

### Wireframe below

![img_5717 1](https://cloud.githubusercontent.com/assets/2292925/8234226/8cb0eebe-15d2-11e5-9442-d687367b8e03.jpg)

tmtmtmtm:
We really need a favicon for the site too — not sure if that will automatically come out of this work —  or should we open a separate ticket for that?

zarino:
@tmtmtmtm: new ticket please!

zarino:
I've started work on this ticket in branch: [issues/279-new-homepage](https://github.com/everypolitician/viewer-sinatra/compare/issues/279-new-homepage?expand=1)

zarino:
Ok, pull request here: https://github.com/everypolitician/viewer-sinatra/pull/303

Things, aside from that pull request, that still need doing:

* Proper text content for the bottom half of the homepage
* Wiring up the search box so it actually takes you to the country pages
* Logo (see #299)
* mySociety colours for buttons etc

zarino:
The pull request (#303) has been merged, and tickets opened for the remaining tasks:

* Explanatory text for the homepage – #308 
* Making the search box take you to country pages – #307 
* Standardising the list of countries in the search box – #306 
* EveryPolitician logo – #299

Issue #170, opened 2015-05-12
Make the Country List page look nicer


tmtmtmtm:
At the moment http://data.everypolitician.org/ is a separate homepage with a list of countries. Ideally we'll consolidate the sites into one (requires #120), but we'll always need _something_ to show which countries we have, and a non-zoomable map on its own isn't enough. 

Unless or until we resolve the wider issue, it's probably worth doing something quite basic to make the http://data.everypolitician.org/ page itself slightly nicer.

tmtmtmtm:
The current plan is to have both the map and the full list on the same page.

PR #303, opened 2015-06-23
New homepage

zarino:
Ok, so we've got a new homepage hero with search box, and a mySociety-compliant typeface, for #279.

You can access the new homepage at `/new_index`.

screen shot 2015-06-23 at 15 28 17

The search box currently searches *all countries in the world* – because I assume we'll want people to be able to see a page for a country even when we don't yet have any data for it. (The search box doesn't actually take you *anywhere* yet – we'll need to work out the best way of doing this.)

screen shot 2015-06-23 at 15 32 59

The text content below the homepage hero is all placeholder content.

I've not changed it to use mySociety colours for buttons yet, because that would be awkward to limit to just the new homepage. It's a quick fix we can do later.

If you're happy for the (unadvertised) `/new_index` to be accessible, then this can all be merged / deployed without affecting the existing pages at all. Even the typeface change is limited only to pages that inherit from the `new_layout.erb` base template.


tmtmtmtm:
@zarino, looking good! Is the plan that this will be the only version of the map we display anywhere? It looks good here, but I'd be concerned if we completely lost the ability to browse that way, especially as we're cutting off chunks of the world from this one. 

What determines the order of the countries returned when you start searching? It's a little surprising for China to come first if you start by typing "P" for example…

I also think the most recent consensus was that although it should look _more_ mySociety-esque than at present, we want it to still stand quite strongly on its own — so I'm not 100% sure about adding the mySociety header-bar. But I'll chat about this some more with @paullenz tomorrow.

zarino:
> Is the plan that this will be the only version of the map we display anywhere?

I had in my head that we were going to put the map at the top of the "browse all countries" page (currently serving as the data.everypolitician.org homepage) – but I now realise we didn't discuss this very much. It makes more sense there, as it's just a visual representation of the plain list below it.

> What determines the order of the countries returned when you start searching?

Currently they're ordered alphabetically. We can turn on "relevancy-sorting" in [the plugin we're using](http://baymard.com/labs/country-selector), which might do something more intuitive. I've not tried it yet though.

tmtmtmtm:
> put the map at the top of the "browse all countries" page 

Yep, that sounds good to me

> Currently they're ordered alphabetically

Not for me, they're not:

screen shot 2015-06-23 at 16 25 55

zarino:
@tmtmtmtm – turns out, the plugin does a bunch of "clever" things I totally didn't expect. [I've removed some of it](https://github.com/everypolitician/viewer-sinatra/commit/bb0a163a0cb56739cbcb6ba8b3cf5662b6d19ced) which was artificially promoting certain "popular" countries (USA, UK, China) over others even when they were less relevant, but the rest of the weirdness is basically the plugin working as expected. We can alter it if you think it's very confusing, but I have a feeling most people won't notice.

**In case you were interested:**

The reason China is scoring so highly for "P" is because one of China's synonyms is "People's Republic", and since the "P" is at the start of a word, it gets sorted into a top group along with other countries with a "P" at the start of their name or their synonym: "French Polynesia", "Korea, People's Republic of", "Pakistan" etc. Then, with all of these top hits having equal relevancy*, the plugin sorts alphabetically on display name, which gives the order you see: "China", then "French Polynesia", "Korea", "Lao", "Portugal"… all the way to "Taiwan, Province of China" at which point it runs out of countries with "P" at the start of one of their names and lists any countries with "P" anywhere in their name(s), starting alphabetically with "Cape Verde".

If we wanted to stop this happening, we could remove the "People's Republic"s from each company's synonyms. But it'll no doubt still happen for other letters. We could stop the autocomplete from returning results until a few more letters have been entered, but that'll make the autocomplete feel slow. Or we can improve the plugin to do the right thing, but I'm not sure how long that would take.

<i>* This is obviously the bit where the plugin gets it wrong – these results don't all have equal relevancy, any human could tell you that.</i>

app.rb
```diff
@@ -18,6 +18,7 @@
   c.delete :legislatures
 
   c[:name] = c[:country]
+  c[:code] = c[:code]
```
tmtmtmtm: Was this really necessary? Or is this a leftover artifact of you playing with something else here at some point?

zarino: Sorry, will delete.

public/javascript/main.js
```diff
@@ -215,4 +215,16 @@ $(function(){
   });
 
   $('html').removeClass('no-js');
+
+  // http://baymard.com/labs/country-selector
+  $('.js-select-to-autocomplete').selectToAutocomplete().on('change', function(){
+    console.log('user selected country:', $(this).val());
```
tmtmtmtm: We should probably remember to remove this before it actually goes live.

tmtmtmtm:
*nod* — that all makes at least some degree of sense. As you say I doubt it'll really matter much to most people. We will, however, definitely need to extend the list though to include places that we cover that aren't in their list (e.g. Kosovo), where we use different variations of the name (Cabo Verde, Congo-Brazzaville), or split out places where we have separate pages that they want to group together (Scotland, Wales, Northern Ireland) — and also remember to keep it in sync as we add new countries (ideally that could be automated — the build process currently fails if we try to create data for a country that doesn't cleanly map to the data in `IsoCountryCodes.for_select`, for example, unless we explicitly provide our own override in a file like https://github.com/everypolitician/everypolitician-data/blob/master/data/Wales/meta.json

The [FAQ](http://baymard.com/labs/country-selector#faq) suggests the best approach would be to feed it the same list that we're using ourselves — though at the minute that list is only for countries we have data for, so we'd want to think about the best way to extend that.

tmtmtmtm:
Merging this. Opened #306, #307, and #308 to track things still to do.

tmtmtmtm:
We need to also move the 'live' map to the country list page (#170)

PR #364, opened 2015-07-07
Add the map to the Data page

### LINKED ISSUE #170: Make the Country List page look nicer

tmtmtmtm:
Once we put the new homepage (#303) live, we'll no longer have the map that's on the existing jekyll homepage. So let's pre-emptively add it to the big list-of-countries page. 

screen shot 2015-07-07 at 15 22 55

Closes #170

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -20,7 +20,7 @@ set :erb, trim: '-'
 get '/' do
   @countries = ALL_COUNTRIES.to_a
   @cjson = cjson
-  erb :front_index
+  erb :front_index, :layout => :new_layout
 end
 
 get '/new_index' do
--- a/views/front_index.erb
+++ b/views/front_index.erb
@@ -1,7 +1,7 @@
 <div class="page-section" id="home">
   
     <div class="container">
-        <div id="vmap" style="width: 100%; height: 400px"></div>
+        <div id="vmap" style="width: 100%; height: 400px; margin-bottom: 2em"></div>
 
         <ul class="grid-list">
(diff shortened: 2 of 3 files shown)


tmtmtmtm:
@zarino Any objections to me simply putting this version live already, and iterating from there? It uses the `:new_layout`, and I'm not sure whether you're happy enough with that yet to actually put something live using it.

My working assumption is that it's better for me to push as much of this stuff live as possible, and then have you clean up the stuff that isn't quite right, than have you do all of this fiddly work. But that actually slows you down, and you'd be happier just implementing everything yourself, that's fine too :)

[NB: You can click the 'deployed' link above to see the staged version of this on heroku]

zarino:
@tmtmtmtm – Go for it!

zarino:
@tmtmtmtm – if anything, I'd add just a bit of margin below the map. `#vmap { margin-bottom: 2em }` should do it.

tmtmtmtm:
Great. At the minute the map styling is just inlined on the div, so I've added it there. That's the sort of thing you can feel free to tidy up later if you have time :)

PR #375, opened 2015-07-09
About pages

davewhiteland:
Broadly these are the words from shared google doc with some (still) outstanding comments not quite addressed. This is being precipitated by me going on leave, sorry :-1:


tmtmtmtm:
Pushing live for now — we'll need to iterate quickly on these, though.

PR #381, opened 2015-07-09
Add a page that lists every country we hope to have

tmtmtmtm:
This isn’t linked to from anywhere, but can be spidered by the process that builds the static site, so as to ensure we get the “No data” page for the countries we don’t have yet.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -74,6 +74,11 @@ get '/about.html' do
   erb :about
 end
 
+get '/status/all_countries.html' do
+  @world = WORLD.to_a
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
@chrismytton : can you check that this looks like a plausible approach, and if so change the process that builds the static site to start from here, rather than /

chrismytton:
@tmtmtmtm I've pushed an extra commit to this branch which changes where the `wget` process starts crawling, so this should be good to go now :+1:

PR #387, opened 2015-07-09
Logo and favicon

### LINKED ISSUE #299: Finalise EveryPolitician logo
@wrightmartin did some very quick sketches, including:

every-p

It needs more work, until it becomes something we're happy with. Then we can implement it in the new website branding (as part of / after https://github.com/everypolitician/viewer-sinatra/issues/279) and in a favicon/avatar (https://github.com/everypolitician/everypolitician/issues/90).

zarino:
Fixes #299.


tmtmtmtm:
:+1: 

I resolved the merge conflict here (clicking the logo now goes to the home page), and also replicated the logo to the LHS at the top of country pages (when the country name goes to the middle). This is all a little jarring, and whilst we can certainly live with it, it's probably worth thinking about a version that has more consistency across different pages.

PR #466, opened 2015-07-31
Display number of politicians

### LINKED ISSUE #429: Add Person count to homepage
We now have a count of People per legislature in `countries.json` — sum these and add it as a running count on the homepage.

tmtmtmtm:
Add a count to the homepage for how many people we have.

This might make the line too long. @zarino, you think this looks OK like this, or should we do something different?

screen shot 2015-07-31 at 15 40 23

Closes #429


zarino:
It's a but long, but not the end of the world.

PR #478, opened 2015-08-04
Resize nav logo on country/term pages.

### LINKED ISSUE #389: Improve visual balance of country page headers
I'm still quite attached to the idea of the EP logo from the homepage being replaced by the active country name, and being centred on the screen (quite unusual). But if we do it, when the EP logo moves to the left-hand side of the header, it needs to get smaller too:

lines

zarino:
Fixes #389.

The left-hand logo now lines up properly with the rest of the text in the header.

screen shot 2015-08-04 at 17 38 51


tmtmtmtm:
:+1:

PR #485, opened 2015-08-06
Improve vertical spacing on short but wide screens

zarino:
Here's a before and after shot, of how a country page looks on a short (800px) but wide screen:

animation

It compresses down even further for even shorter screens (below 600px high).


tmtmtmtm:
Great, thanks @zarino!

PR #490, opened 2015-08-07
Add nav to docs

davewhiteland:
this has added a couple of column-wrangling divs to each of the docs pages, which has changed the indenting on the body of the texts, but I don't think there are material changes in the words :-)

### DIFF
--- a/views/about.erb
+++ b/views/about.erb
@@ -1,91 +1,95 @@
 <div class="container" id="about">
   <div class="page-section">
+    <div class="row">
+      <div class="column-one-quarter">
+         <%= erb :about_nav_menu %>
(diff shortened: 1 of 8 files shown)


tmtmtmtm:
screen shot 2015-08-07 at 16 31 07

tmtmtmtm:
@davewhiteland was dropping the "the" in "Getting most recent data" deliberate?

davewhiteland:
yup, I suspect the line might be too long with it, but that's a bit hand-wavy how-long-is-a-piece-of-[in-this-font]-string

tmtmtmtm:
Looks great then! :shipit:

PR #533, opened 2015-08-27
add how-to-use-data page, w. Gender Balance example

davewhiteland:
please check for truthiness

Last week I was in a spin about the image proxying, but really I don't think it's terribly important. It's basically a refinement. Can add it later if you think it's very important.


tmtmtmtm:
No, we definitely shouldn't include the image proxy stuff — it's a misleading distraction from the core idea.

tmtmtmtm:
@davewhiteland is this obsolete now, or do we want to rescue any of it?

PR #574, opened 2015-09-07
Issues/421 how to scrape

davewhiteland:
guidance for contributing and writing scrapers ready for GLOW.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -120,6 +120,10 @@ get '/data_structure.html' do
   erb :data_structure
 end
 
+get '/scrapers.html' do
+  erb :scrapers
+end
+
 get '/*.css' do |filename|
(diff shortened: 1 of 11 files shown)


tmtmtmtm:
@davewhiteland Thanks a bunch — this is great! I've made a few tweaks, and the only thing I can really think off that's important but missing is an explanation of when you might want to use a composite column group on the `save` (i.e. there will be more than one record per person, such as if you're scraping multiple terms). If you're happy with the current version, though, I'm OK with putting this live as-is, and then we can add that separately.

davewhiteland:
Yup, have raised #577 to add the explanation about composite columns

davewhiteland:
managed to not automatically close this, but it's merged.

PR #620, opened 2015-09-09
Update installation instructions for this fork

andylolz:


tmtmtmtm:
Thanks @andylolz. (I'm not sure why Travis is failing on that test, but I'm assuming it's a transitory thing as the other one passed)

andylolz:
Yeah, weird. All tests seem to pass, but non-zero exit code.

dracos:
Pull requests from external repositories do not have access to secure information (as they could then find that out by adding a `printenv` and submitting a PR). Looks like `scripts/release.sh` assumes it can see those keys and dies at the SSH key bit.

andylolz:
Ahh gotcha, thanks @dracos. PR sent that possibly helps?

tmtmtmtm:
Thanks @andylolz. I'm going to leave this one for @chrismytton to look at next week, in case there are any implications here that I'm not aware of.

andylolz:
Yeah, good plan. I don’t really know what `scripts/release.sh` is doing, so it’s just a guess on my part really.

Sorry to keep distracting you from the business of scraping!

PR #625, opened 2015-09-10
Stop Travis always failing for external PRs

andylolz:
Refs https://github.com/everypolitician/viewer-sinatra/pull/620#issuecomment-139165313

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -38,8 +38,8 @@ update_politician_image_proxy() {
 }
 
 main() {
-  add_ssh_key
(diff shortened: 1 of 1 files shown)


andylolz:
Tests pass so I guess this works.

dracos:
`add_ssh_key` was moved outside the `if` in 721cffb811309e6bf41ab7e113794218d247ba15 along with `update_politician_image_proxy`; the latter was moved back inside in eb981ccd2956848a3aa6b87a939915c7ef6e2537 but `add_ssh_key` left outside, so looks good to me.

tmtmtmtm:
Thanks @dracos 

Let's go for it then…

PR #662, opened 2015-09-14
Only use flex if flex-wrap is supported.

### LINKED ISSUE #127: Flexbox wrapping doesn’t work in Safari
None of the wrapping declarations are being accepted by Safari.

This causes the person/party lists to just extend off the screen.

screen shot 2015-05-01 at 16 17 51

First port of call: check that `-webkit-box-wrap` is the right thing to use, and that `wrap` is a valid value for it.

dracos:
If flex is supported and flex-wrap is not, you get one long horizontal
list spilling out to the side of the browser. Fixes #127.


tmtmtmtm:
Thanks @dracos 

@zarino this look good to you?

dracos:
I'll just make a smaller modernizr.js file.

zarino:
@dracos Looks good, although the `@include .unstyled-list` on [line 321](https://github.com/everypolitician/viewer-sinatra/blob/4c18617d2e59e4a8a884f086bf25c6efd3fffa10/views/sass/_components.scss#L321) needs to be moved outside the `.flex-wrap` modernizr check (it should be applied regardless of whether the browser supports flexbox or not).

dracos:
@zarino Doh, of course, thanks. Fixup commit done, and modernizr JS shrunk from 4225 to 533 bytes!

tmtmtmtm:
Yay. Thanks @dracos

PR #668, opened 2015-09-14
Stop Safari redirecting to localhost on page blur.

dracos:
An onchange is fired as the field had focus, so don't redirect if we
don't have a selected entry.


tmtmtmtm:
Well found! :rocket:

PR #767, opened 2015-10-09
Move JSON download button to avoid confusion over scope

### LINKED ISSUE #370: Split the Download buttons
Currently we have download buttons on the Term Table page for both CSV and JSON:

screen shot 2015-07-08 at 10 26 20

However, these are are for conceptually different things: the CSV is only for the term you're currently viewing, whereas the JSON is for the legislature as a whole. 

If we had a separate page for each legislature, then it would make sense to put the JSON download button at the bottom of that (similarly to the one to `countries.json` on the list of countries page), but currently we just group all of them on the country page (as right now most countries only have a single legislature listed anyway). So we should probably move the JSON download to there.

zarino:
Previously JSON download was presented alongside CSV download for a legislature’s membership table, despite the two files containing very different data.

This moves the JSON download link to the top of the term page (and names it clearly with the number of terms it contains) and also adds JSON download links to the country page.

@tmtmtmtm Before this can be merged, you need to pass the correct JSON file URL for each legislature into `country.erb` (in place of `@example_json_url`).

Fixes #370.

screen shot 2015-10-09 at 12 59 07

screen shot 2015-10-09 at 12 59 35

### DIFF
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -1,9 +1,19 @@
 <div class="page-section" id="term">
-    <div class="container">
-        <h1 class="text-center"><span class="avatar"><i class="fa fa-university"></i></span> <%= @term[:name] %></h1>
+    <div class="container text-center">
+        <h1><span class="avatar"><i class="fa fa-university"></i></span> <%= @term[:name] %></h1>
+
       <% unless @term[:start_date].to_s.empty? and @term[:end_date].to_s.empty? %>
-        <p class="text-center"><%= @term[:start_date] %> – <%= @term[:end_date] or 'current' %></p>
+        <p><%= @term[:start_date] %> – <%= @term[:end_date] or 'current' %></p>
       <% end %>
+
+        <a href="<%= @urls[:csv] %>" class="term-download">
+          <i class="fa fa-download"></i>
+          <% if @terms.length > 1 %>
+            Download all <%= @terms.length %> terms as JSON
+          <% else %>
+            Download as JSON
+      <% end %>
+        </a>
     </div>
 </div>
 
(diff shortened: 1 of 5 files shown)


views/term_table.erb
```diff
@@ -1,9 +1,19 @@
...
-        <h1 class="text-center"><span class="avatar"><i class="fa fa-university"></i></span> <%= @term[:name] %></h1>
+    <div class="container text-center">
+        <h1><span class="avatar"><i class="fa fa-university"></i></span> <%= @term[:name] %></h1>
+
       <% unless @term[:start_date].to_s.empty? and @term[:end_date].to_s.empty? %>
-        <p class="text-center"><%= @term[:start_date] %> – <%= @term[:end_date] or 'current' %></p>
+        <p><%= @term[:start_date] %> – <%= @term[:end_date] or 'current' %></p>
+      <% end %>
+
+        <a href="<%= @urls[:csv] %>" class="term-download">
```
tmtmtmtm: Presumably this was meant to be a link to the JSON rather than the CSV? Either way, I think even having this link here at all is confusing for not really much value, especially now that you can download the full JSON from the country page. I think I'll just remove it…

tmtmtmtm:
Thanks @zarino. The template already gets enough information from `countries.json` to construct the URL for the downloads, so I was able to just use that.

I've also harmonised all the other download buttons to use this new approach of explicitly saying "Download as…" on large screens. Happy to back that out again if there was a reason you were only doing it for the single case.

zarino:
> I've also harmonised all the other download buttons to use this new approach of explicitly saying "Download as…" on large screens

That's great!

PR #769, opened 2015-10-09
Show politician counts in country autocomplete

### LINKED ISSUE #386: Visually distinguish countries we have data for on homepage search box
The new search box should highlight the countries we do actually have information for, rather than the user having to go through to a page that says "Sorry, nothing here".  (Though we should still have that page, so that we can explain how people can help, etc.)

zarino:
Rather than hacking directly into the select-to-autocomplete module, we tweak the jQuery autocomplete settings *after* it has been instantiated on the page, supplying our own custom `_renderItem` method, which uses a list of people counts pulled in from the `title` attributes of the original select options.

Fixes #386 by showing people which countries have data and which don’t.

screen shot 2015-10-09 at 16 42 59

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -27,8 +27,6 @@ set :erb, trim: '-'
 get '/' do
   @countries = ALL_COUNTRIES.to_a
   @person_count = @countries.map { |c| c[:legislatures].map { |l| l[:person_count].to_i  } }.flatten.inject(:+)
-    .to_s.reverse.gsub(/(\d{3})(?=\d)/, '\\1,').reverse
-    # http://stackoverflow.com/questions/1078347/is-there-a-rails-trick-to-adding-commas-to-large-numbers
   @world = WORLD.to_a
 
   @world.each { |slug, country|
--- a/app.rb
+++ b/app.rb
@@ -30,6 +30,10 @@ get '/' do
     .to_s.reverse.gsub(/(\d{3})(?=\d)/, '\\1,').reverse
     # http://stackoverflow.com/questions/1078347/is-there-a-rails-trick-to-adding-commas-to-large-numbers
   @world = WORLD.to_a
+
+  # :TODO: Tony, replace this with something that uses real person_count.
+  @world.each { |slug, country| country[:totalPeople] = [0, 1, 20, 98765].sample }
+
   erb :homepage
 end
 
--- a/app.rb
+++ b/app.rb
@@ -31,8 +31,10 @@ get '/' do
     # http://stackoverflow.com/questions/1078347/is-there-a-rails-trick-to-adding-commas-to-large-numbers
   @world = WORLD.to_a
 
-  # :TODO: Tony, replace this with something that uses real person_count.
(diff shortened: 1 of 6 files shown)


zarino:
@tmtmtmtm As mentioned in Slack, you'll need to replace the placeholder line in `app.rb`, to pass in the actual country person counts.

tmtmtmtm:
@zarino — Yay! I've added a real count there, and factored out the commification code, so we can use that too. Everything now looks good:

good screen shot 2015-10-09 at 18 24 14

_except_ for the box growing overlong when you first start typing:

overlong screen shot 2015-10-09 at 18 23 18

(The current version doesn't do that):

current screen shot 2015-10-09 at 18 24 58


I think this is a big improvement over the current version, so I'm putting it live anyway, and if it's complicated to fix, I'm not sure it's really worth it. But if it's simple, it's probably worth adding.

Thanks!

zarino:
> except for the box growing overlong when you first start typing:

Yeah, I spotted that. I think it's down to how the javascript decides how wide to make the list items (eg: it's not factoring in the extra padding on the right, which means the list items end up wider than they should be) but I haven't spent much time testing that.

> if it's complicated to fix, I'm not sure it's really worth it.

I think that's probably the case. But I'll allocate half an hour this morning to investigating what can be done. I agree that it's a bit disconcerting when you first search, and all the alignment seems broken for a second or so.

tmtmtmtm:
It only ever seems to happen when you've only typed a single letter. As a fall-back solution, perhaps just don't start doing the auto-complete until at least two letters have been typed?

zarino:
> It only ever seems to happen when you've only typed a single letter.

It happens whenever there's a really long country name that matches what you've typed. It looks like the widest names is "Congo-Kinshasa (Democratic Republic)", with "Sahrawi Arab Democratic Republic" and "Saint Vincent and the Grenadines" close runners up.

screen shot 2015-10-12 at 09 55 28

So if your first letter is one of the letters from either of those names, you get a wide menu.

I'd be wary of delaying the autocomplete until you've typed more letters, because you'll lose the instant feedback (compare to search boxes elsewhere on the web: Google, Facebook, Chrome, Twitter – they all update instantly for each letter you type).

zarino:
Oh yeah, forgot to say, the "long names cause the menu to overflow" issue happened *before* this PR was merged too – it's just that we didn't notice it so much:

screen shot 2015-10-12 at 10 21 39

screen shot 2015-10-12 at 10 20 52

zarino:
Aha! Turns out this is a well known issue, and there's a [simple fix](http://stackoverflow.com/a/11845718/3096375), which I've pushed to https://github.com/everypolitician/viewer-sinatra/pull/781. It results in a box that looks like this:

fixed

So the long place names end up getting wrapped onto multiple lines. If you're happy with that, [merge the pull request](https://github.com/everypolitician/viewer-sinatra/pull/781).

Otherwise, if you don't want the place names to wrap on desktops (they'll still wrap on narrow screens) we could make the search box wider:

wide-box

Or just make it wider when it receives focus (a couple of sites do this—including our own TWFY—although usually only in nav bars, where space is at a premium):

expanding-box

Issue #721, opened 2015-09-30
Tidy the 'countries' page


tmtmtmtm:
http://everypolitician.org/countries.html

Now that we have almost every country that is actually visible on the map, we should rethink whether it makes as much sense any more (or if we should more prominently highlight the ones we're missing, possibly combining it with [/needed.html](http://everypolitician.org/countries.html) in some way?)

Is a massive columned list the best way to display the countries now that there are more than 200 of them?

Should this page also have the search box that's on the home page? 

Should it even exist at all?

zarino:
I think the countries page is still useful, but as you say, it needs some changes now that it's matured.

Off the top of my head:

* Change the map so that it highlights missing countries more strongly. Use the EP salmon colour perhaps? Need to check whether that sits well with green for existing countries, or whether the existing countries need to be scaled back to grey.
* Combining this page with http://everypolitician.org/needed seems sensible. Especially since the map will now be more clearly showing both.
* To combat the "gigantic list of countries" effect, maybe split them into continents?
* I'm not convinced the little people in circles is adding anything to such a long list – feels a bit like noise to me. Maybe try without the circles? All the country names could also easily be a bit smaller, and no longer bold, to reduce some of the plonky weight.

tmtmtmtm:
I'm open to all sorts of options on this, though I'm a little worried by the geo-politicial minefield edge-cases involved in splitting into continents…

zarino:
@tmtmtmtm Have you come across any less minefieldy ways of splitting up a list of countries, because I'm drawing a blank.

I guess you could split them into little alphabetical groups, so all the "A" countries, then all the "B" countries, etc. At least then you could quickly scan down to where you guess your country might be.

tmtmtmtm:
Grouping by first letter is probably OK. Maybe we should add the filter box from the homepage here again too?

tmtmtmtm:
Closed by #770

Issue #722, opened 2015-09-30
Remove GLOW banner?


tmtmtmtm:
We should probably remove this at some point quite soon, though it would still be nice to mention _something_ about it somewhere.

zarino:
Well, even if we get rid of the one on the homepage, there's still the same banner on the "Contribute" page – we could leave that there?

![screen shot 2015-10-08 at 16 51 45](https://cloud.githubusercontent.com/assets/739624/10371820/02200cc2-6ddd-11e5-9d8c-54a29b7ef89d.png)

tmtmtmtm:
I guess so, though even there it seems disproportionately prominent now…

zarino:
For now I’m just going to remove the banners. You can decide with @MyfanwyNixon as to how you want to work a mention of GLOW into the site – if you still do.

MyfanwyNixon:
Personally I think the moment has passed now!

PR #770, opened 2015-10-09
Split countries page into alphabetical sections

zarino:
Also moves the JSON download link further up the page, and colours missing countries red on the map so that they stand out.

The most minimal solution to #721. (Given more time, the next step would be to add a search box to the top of the page.)

screen shot 2015-10-09 at 17 42 28


tmtmtmtm:
Thanks @zarino — looking good. I wasn't completely convinced by having the download link be quite so prominent at the top, so I've moved it down to the very bottom for now. But I'm open to persuasion on putting it back up top…

Making the missing countries be red works well, and instantly made me realise that clicking on those should take you to the `/needed` page. I'll add that separately. (#772)

zarino:
Yeah, I wasn't sure how prominent you wanted the JSON download to be (arguably, now you've got a lot more countries, it's more useful). But I mostly moved it to match the JSON download link on the individual term pages. Since you've removed it from them (I think) you might as well not have it here either.

Incidentally, every time I looked at the page, I sort of expected that red link to say something like "…and 24 left: can you help?" or something. Because the link is the same colour as the missing countries on the map, and there's a nice balance between saying what you've got and what you still need. But I appreciate you probably don't have that number to hand.

Even with #722, I still feel like there needs to be a clear link from the *Countries* page to the *Needed* page. Or they need to be combined.

tmtmtmtm:
> "…and 24 left: can you help?"

Oh, that's a nice idea… I'm sure I can find a way to make that work!

PR #781, opened 2015-10-12
Correctly limit width of homepage autocomplete dropdown

zarino:
Fixes a styling bug related to #769, where wide country names made the dropdown overflow the width of the search box above it.

fixed

Also, since menu items are more likely to wrap onto multiple lines now, we increase the line-height slightly to make them look better when they do wrap.

And for consistency, we use the same prototype-extending paradigm for the other autocomplete hack we added in 338648794956f315b7adca356caa8b6b99e7d792.

### DIFF
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -227,6 +227,13 @@ $(function(){
     $(this).next().trigger("focus");
   });
 
+  // Fix the incorrect default autocomplete width, which meant that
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
:+1: Looks good to me! I don't think there's a problem with long names being split over multiple lines, though if you want to make the box longer later, I'm certainly OK with that too.

PR #791, opened 2015-10-14
Make sort_name the default sort

tmtmtmtm:
By default we should order the names according the sort_name (where available — where it's not, it'll fall back on the `name` field).

<s>This has a slightly odd side-effect of meaning that once someone uses the clicky-arrows to re-sort, there's no way of returning to this order.  There may be a way of doing something with `restoreOriginalSortOrder` in the Javascript here, but it's not immediately obvious what that should be.</s>

This means that the behviour is:
* default: sort by sort-name (usually last name first)
* click on 'sort' in name field: sort alphabetically by display name (usually given name first)
* click on 'sort' in name field again: sort reverse alphabetically by display name
* click on 'sort' in name field again: return to sorted by sort-name

(With no way to reverse sort by sort-name)

@zarino any thoughts or comments on what you'd expect to happen here (and if that's different from what currently happens, what might be involved in making that work)? Is what we have now enough of an improvement for cases where we _have_ sort_order (try Slovakia for an example), or is it likely to be too confusing?

### DIFF
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -131,10 +131,16 @@
       $table.data('unrowspanned', true);
     }
 
+    var getSortValue = function($element){
+      // Table cells can specify a custom value by which
+      // they will be sorted (useful for names and dates).
+      return $element.attr('data-sortable-sortvalue') || $element.text();
+    }
+
     var sortRows = function sortRows($tbody, $trs, columnIndex, sortOrder){
       $trs.detach().sort(function(rowA, rowB){
-        var valueA = $(rowA).children('td').eq(columnIndex).text();
-        var valueB = $(rowB).children('td').eq(columnIndex).text();
+        var valueA = getSortValue( $(rowA).children('td').eq(columnIndex) );
+        var valueB = getSortValue( $(rowB).children('td').eq(columnIndex) );
 
         var moveUp = 1;
         var moveDown = -1;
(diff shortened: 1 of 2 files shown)


zarino:
So the table is initially sorted (A-Z by "sort_name" which might be last name, or might not, depending on the country) on the name column.

This means the name column acts very differently to the other columns (which start off unsorted, and have three states in total: unsorted, sorted A-Z, sorted Z-A).

I think our `sortable` plugin needs a new concept for this "already sorted" type of column. It should:

* Show the column as sorted (little downward-pointing blue arrow) on pageload.
* Alternate between *only two states* when the header is clicked: sorted ascending, and sorted descending. The column cannot be unsorted.

I estimate it'll take about half a day to make the necessary changes to the javascript code.

zarino:
@tmtmtmtm: I've done what I think is the most intuitive thing. The "Name" column is shown as sorted on pageload, and the second click on the header takes it back to being sorted A-Z, rather than attempting to unsort it. (Meaning the "Name" column can only ever alternate between A-Z and Z-A sorting.)

sort

It's all done via the `data-sortable-presorted` attribute, which is now documented at the top of the `$.sortable` plugin definition.

tmtmtmtm:
@zarino — looks good to me! Thanks :rocket:

PR #822, opened 2015-10-22
'source' is a required field on submitted data

### LINKED ISSUE #683: Document the need for a `source` column
All rows should include a `source` URL

davewhiteland:
closes #683


tmtmtmtm:
I think we really need to add this to the example table as well — I suspect a lot of people will largely take their steer from that, and look more to the text for explanation if needed.

tmtmtmtm:
:+1:

PR #883, opened 2015-10-28
Second attempt at useful description of how to use the data

davewhiteland:
This now includes clear (I hope) example of how to construct URLs from countries.json


views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			This page is for you if you&rsquo;re a developer who is planning on
+			building something wonderful or brilliant with the data we have on your
+			country&rsquo;s politicians. It tells you how to get the data and, broadly,
+			how it&rsquo;s structured (spoiler: CSV or Popolo JSON).
+        </p>
+		<p>
+			We believe that by making this data available in consistent and
+			useful formats, if you write something neat that uses, for example, the data
+			for your country&rsquo;s politicians, and open it up under a free software open
+			source license... then other people will be able to adapt what you&rsquo;ve
```
tmtmtmtm: covering all the bases?

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			done to run in their country too. That works because their data will be
+			formatted the same way that yours is: that&rsquo;s our side of the deal.
+		</p>
+		<h2>
+			How to find what data we&rsquo;ve got for you
+		</h2>
+		<p>
+			Humans look at the <a href="/">map on the front page</a>. But
```
tmtmtmtm: The map on the front page isn't really a useful/usable one, though… the interactive map is actually on /countries.html

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			you&rsquo;re a developer: you look in
+			<code><a href="https://raw.githubusercontent.com/everypolitician/everypolitician-data/master/countries.json">countries.json</a></code>.
+		</p>
+		<p>
+			That&rsquo;s because <code>countries.json</code> is effectively
+			EveryPolitician&rsquo;s index. It contains the complete set of countries we
+			have data for as a JSON array of country objects. Those contain identifying
```
tmtmtmtm: I tripped on this on first read, as I grouped it as "countries we have (data for as a JSON array of country objects)" rather than "(countries we have data for) as a JSON array". Might just be a matter of a comma after the 'for'?

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			fields (like name and country code) but also the SHAs of the git commits of
+			the most recent versions of each legislature within those countries, and the
+			paths you need to <em>find</em> them. From <code>countries.json</code>, you
```
tmtmtmtm: This line confuses me slightly, as it's not clear to me what the "them" refers to.

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			can construct the URLs of the data you want.
+		</p>
+		<p>
+			Every time we change any of the data, we update <code>countries.json</code>.
+			See <a href="repo_structure.html">getting the most recent data</a>
+			for details about using this crucial file.
+		</p>
+		<h2>
+			Four policies for <em>getting the data</em>
```
tmtmtmtm: Maybe "approaches to" would work better?

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			can run the same process.
+		</p>
+		<p>
+			It depends on your project, but really we&rsquo;re suggesting you
+			write an import or update script that populates your database; or make
+			sure your code can run from scratch against a fresh download. This means you
+			need to consider things like: what happens if a politician&rsquo;s name changes?
+			Someone is removed? A whole new government is elected? Make sure your
+			application can handle this change.
+		</p>
```
tmtmtmtm: We probably need to say something about EveryPolitician IDs here — though we have the slight problem that that's about to change quite radically. However, whether on the current or hopefully-about-to-become-current approach, it's something people will need to be able to handle — either IDs aren't stable at all, or they'll hopefully be mostly stable, but will still be subject to change (e.g. if we discover two records with different IDs are really the same person and need to merge them.)

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+		<p>
+			We used this event-triggered mechanism for our own example application <a
+			href="http://www.gender-balance.org/">Gender Balance</a>. We started running
+			Gender Balance when there were still a lot of countries missing from the
+			EveryPolitician data set. But because Gender Balance uses this event-driven
+			approach, it pulls in the latest changes whenever the data is updated
+			&mdash; basically, it keeps itself up-to-date. So whether there&rsquo;s an
+			election in Sweden or a politician in Thailand changes their name, Gender
+			Balance will incorporate the change it as soon as EveryPolitician has the
```
tmtmtmtm: <s>it</s>

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			grab the JSON data and use your favourite JSON library to jump right into
+			having structured data in your application.
+		</p>
+    <a name="#rawgit"></a>
+		<h2>
+			Where to get it from: <code>rawgit.com</code>
+		</h2>
+		<p>
+			Although the EveryPolitician data is published on GitHub, your code
+			should&rsquo;t try to download it from there. You can get away with that as
```
tmtmtmtm: should't → shouldn't

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+			a human, but if your application tries it, it will run into problems.
```
tmtmtmtm: This isn't really true. Lots of apps will be able to consume the data just fine directly from GitHub. Most server-side consumption will work without any problem. The issue is really if you want it client-side, in JavaScript.

views/use_the_data.erb
```diff
@@ -0,0 +1,256 @@
...
+		</p>
+		<p>
+			Because we &ldquo;publish&rdquo; our data on GitHub, it&rsquo;s
```
tmtmtmtm: Why the scare-quotes?

tmtmtmtm:
This is coming on well, though there's quite a bit of handwaving about the webhooks bit for something that we're recommending as the best method. I think this needs to be a lot clearer on *how* to do this — especially as that'll most likely involve them using the app manager (as I don't think you can set up webhooks directly on a repo that isn't your own…)

views/use_the_data.erb
davewhiteland: Yeah it's clumsy this... problem is that just making a tool that works with the data doesn't mean anyone else can use it. Gah.

views/use_the_data.erb
davewhiteland: Excellent, ta

views/use_the_data.erb
davewhiteland: Better: "It contains a simple JSON array of objects, one for every country we've got data for."

views/use_the_data.erb
davewhiteland: Clarified by dropping some of the detail (e.g., about the SHAs: that's described in the part where we actually *use* them) so just say "metadata", cos that's what it is --

"These contain identifying fields (like name and country code) as well as metadata for each legislature within those countries. From countries.json you can construct the URLs of the files that contain the data you want."

views/use_the_data.erb
davewhiteland: Removed; this was based on me earlier saying "get it from rawgit.com" so in that sense it doesn't feel like publishing is the right word for putting it on github; but that's me being overcautious about the use of rawgit -- as you say, that's really for the specific case of client consumption, so it's not unreasonable for someone to get data from github, if that works for them, right?

views/repo_structure.erb
```diff
@@ -11,16 +11,101 @@
...
+          &nbsp;&nbsp;
+          /<span class="url-part-sha">commit-SHA</span><br>
+          &nbsp;&nbsp;
+          /<span class="url-part-file">path-to-data-file</span></code>
+        </p>
+        <p>
+          You might expect the domain to be <code>http://github.com/</code>, but for
+          your application <a href="/use_the_data.html#rawgit">we recommend you use <code><span class="url-part-domain">http://rawgit.com</span></code></a>
+          (or for best performance: <code><span class="url-part-domain">http://cdn.rawgit.com</span></code>).
+        </p>
```
tmtmtmtm: I don't think we should ever direct people to "rawgit.com" directly (rather than cdn.rawgit.com). As per the "When people misuse rawgit.com, it costs me money. Please don't be a jerk" part of the [FAQ](http://rawgit.com/faq) you link to in the other section, using the cdn.rawgit version isn't just a "for best performance" nicety — so it should probably be the _only_ version we suggest.

PR #979, opened 2015-11-13
describe use of identifier__ ids in import

### LINKED ISSUE #963: Document incoming identifiers
http://everypolitician.org/submitting.html says that an `id` field is required, but not that it will be turned into a 'legacy' id, in favour of a UUID. We should explain this, and that if you're passing through an ID from elsewhere that you'd like to refer to again somewhere else, you should give it an `identifier__` prefix.

davewhiteland:
For @tmtmtmtm to check please!

closes #963

### DIFF
--- a/views/submitting.erb
+++ b/views/submitting.erb
@@ -18,12 +18,57 @@
           <a href="scrapers.html">writing a scraper</a> for us.
           </p>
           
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
Thanks @davewhiteland :+1:

PR #1175, opened 2015-11-23
new page to summarise data available

### LINKED ISSUE #948: document "what's in the data?"
This came up a lot with the people we talked to at MozFest -- need to document clearly what data we do have (e.g., names, names in different languages, twitter handles, dates, etc)

davewhiteland:
new page answering most common question (imo) from MozFest; closes #948

### DIFF
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -236,12 +236,12 @@
           legislature, which means the data for politicians in the current
           term. Often — but not always — this is also the data we’re most
           likely to be able to source online, so if you see a country only has
-          one term in its data, it may well be the current one. We’re always
-          interested in adding historic data too (that is, data for previous
-          terms), so if you know where we can get it for your country, please
-          let us know. Similarly, if we are lacking the current term (maybe
-          there’s just been an election?), please let us know — it might be
-          that our source has been updated and we haven’t pulled down the
+          one term in its data, it will probably be the most recent one. We’re
+          always interested in adding historic data too (that is, data for
+          previous terms), so if you know where we can get it for your country,
+          please let us know. Similarly, if we are lacking the current term
+          (maybe there’s just been an election?), please let us know — it might
+          be that our source has been updated and we haven’t pulled down the
           latest changes.
         </p>
         <h2>
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -10,15 +10,15 @@
           Short answer: names. And dates of birth, twitter handles, politicial
           group memberships, email addresses, honorific titles, image URLs, and
           all sorts of useful things — all provided we can find them. How
-          complete this data is depends on the sources you and people like you
-          have found for us for each country.
+          complete this data is depends on the sources that people around the
+          world have found for us for each country.
         </p>
         <p>
-          This is a non-technical overview. To properly understand what
-          EveryPolitician offers, read about the <a href="technical.html">two
-          data formats available</a>, and how the <a
-          href="data_structure.html">data structure</a> encapsulates
-          <strong>entities</strong> and <strong>relationships</strong>.
+          This is an overview of what you can find in our data. You might also
+          want to read about the <a href="technical.html">two data formats
+          available</a>, and how the <a href="data_structure.html">data
+          structure</a> encapsulates <strong>entities</strong> and
+          <strong>relationships</strong>.
         </p>
         <p>
           The CSV files, which are often easier to work with, contain a useful
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -160,10 +160,6 @@
                         <td>id for area</td>
                       </tr>
                       <tr>
-                        <th>chamber<span class="superscript">&dagger;</span></th>
-                        <td>name of chamber <code>House of Representatives</code></td>
-                      </tr>
-                      <tr>
                         <th>term<span class="superscript">&dagger;</span></th>
                         <td>id or index for term</td>
                       </tr>
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -10,8 +10,8 @@
           Short answer: names. And dates of birth, twitter handles, politicial
           group memberships, email addresses, honorific titles, image URLs, and
           all sorts of useful things — all provided we can find them. How
-          complete this data is depends on the sources we’ve found for each
-          country.
+          complete this data is depends on the sources you and people like you
+          have found for us for each country.
         </p>
         <p>
           This is a non-technical overview. To properly understand what
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -211,8 +211,8 @@
           For every country, we break the data down into
           <strong>legislatures</strong> (for example, the “parliament” or
           “congress” of the country). Currently, we’re only working with
-          national, that is, top-level, legislatures (as opposed to local or
-          state-level legislatures). In the future we might include those too.
+          national, that is, top-level, legislatures. In the future we might
+          include local or state-level legislatures too.
         </p>
         <p>
           Many countries have a single legislature. Some have two. For example,
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -191,18 +191,19 @@
         </div>
         <p>
           <span class="superscript">&dagger;</span>
-          You can use fields marked &dagger; as IDs, although we can’t
-          guarantee they’ll always be unique <em>outside</em> this legislature.
-          For example, all politicians in this legislature with the same
+          You can consider fields marked &dagger; as identifiers that are
+          unique within this legislature (unlike <code>id</code> itself, which
+          identifies the politician within <em>all</em> of EveryPolitician’s data). For
+          example, all politicians in this legislature with the same
           <code>area_id</code> are representing the same area (which is what
           you’d expect). Where appropriate, the actual value might be a useful
-          value. For example, we like to use <a
+          value. For example, for <code>area_id</code> we like to use <a
           href="http://opencivicdata.readthedocs.org/en/latest/ocdids.html">Open
-           Civic Data IDs</a> like this:
+          Civic Data IDs</a> like this:
           <code>ocd-division/country:us/state:ca/cd:47</code>, but if no such
           mapping is available from our sources, we’ll probably use a slug like
-          <code>area-foo-county</code>. If you can suggest better values whenever
-          you find us using our own ones, please get in touch!
+          <code>area-foo-county</code>. If you can suggest better values
+          whenever you find us using our own ones, please get in touch!
         </p>
         <h2>
           Legislatures and terms
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -346,8 +346,8 @@
         <p>
           Some legislatures have adopted the good practice of assigning unique
           IDs to their politicians, which can be useful when using their
-          official APIs. Whenever we can find IDs like this, we’ll add them as
-          identifiers in this way.
+          official APIs. If you know that your country’s legistature provides
+          such IDs please tell us, and we’ll add them as identifiers in this way.
         </p>
       </div>
     </div>
--- a/views/data_summary.erb
+++ b/views/data_summary.erb
@@ -332,10 +332,10 @@
 ]
 </pre>
         <p>
-          (Incidentally, you can view WikiData entries like this
-          <a href="https://www.wikidata.org/wiki/Q42">wikidata.org/wiki/Q42</a>,
-          or using the Reasonator: in this example, it’s the
-          <a href="https://tools.wmflabs.org/reasonator/?q=Q1091494">Parliament of Åland</a>).
+          (Incidentally, you can view WikiData entries like this:
+          <a href="https://www.wikidata.org/wiki/QQ1091494">wikidata.org/wiki/Q1091494</a>,
+          or using <a href="https://tools.wmflabs.org/reasonator/?q=Q1091494">the Reasonator</a>.
+          This example is the Parliament of Åland.)
         <p>
           We do this for other entities (such as legislatures themselves) as
           well as politicians. Remember that external data sets might not be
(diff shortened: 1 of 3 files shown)


views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+      <div class="column-one-quarter">
+         <%= erb :about_nav_menu %>
+      </div>
+      <div class="column-three-quarters">
+        <h1>What’s in EveryPolitician’s data?</h1>
+        <p class="section-intro">
+          Short answer: names. And dates of birth, twitter handles, politicial
+          group memberships, email addresses, honorific titles, image URLs, and
+          all sorts of useful things — all provided we can find them. How
+          complete this data is depends on the sources we’ve found for each
```
tmtmtmtm: I'm wondering if this should be turned around slightly to make it less "what *we* have found" and more "what *you* have told us about"?

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          country.
+        </p>
+        <p>
+          This is a non-technical overview. To properly understand what
```
tmtmtmtm: You say this is the non-technical overview, but it then immediately starts getting quite technical. The description here of how it all hangs together is good, but it seems likely to immediately be off-putting to a non-tech person just wanting to know whether it's even worth looking into using the data (or having a more technical person look into it)

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+                        <th>area</th>
+                        <td>name of area represented <code>Example County</code></td>
+                      </tr>
+                      <tr>
+                        <th>area_id<span class="superscript">&dagger;</span></th>
+                        <td>id for area</td>
+                      </tr>
+                      <tr>
+                        <th>chamber<span class="superscript">&dagger;</span></th>
+                        <td>name of chamber <code>House of Representatives</code></td>
```
tmtmtmtm: I'm not sure I'd include this. As the file only exists at the level of a single legislature, the value will be the same in every row here. At some point (soon?) we'll completely remove this. If this were the formal documentation for the file formats, we should probably include it and note that it's deprecated, but this doesn't seem like the right place for that…

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+        <p>
+          <span class="superscript">&dagger;</span>
+          Values of fields marked &dagger; will be consistent throughout the
+          legislature: if their values are equal (string comparison), then
+          they’re the same thing. So all politicians with the same
+          <code>area_id</code>, for example, are representing the same area.
```
tmtmtmtm: This is a slightly confusing description. It seems like it's trying to counter some other assumption, but without stating what that is. It's unclear how two `start_date`s that are the same string value refer to two different things, for example, or even two `area` fields.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          Where appropriate, the actual value might be a useful value. For
+          example, where we can, we use <a
```
tmtmtmtm: "where we can" is a little strong. We'd _like_ to use them in more places, but we don't do it very often yet (though feel free to start adding them for other places, or encouraging other people to!)

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        </p>
+        <h2>
+          Legislatures and terms
+        </h2>
+        <p>
+          For every country, we break the data down into
+          <strong>legislatures</strong> (for example, the “parliament” or
+          “congress” of the country). Currently, we’re only working with
+          national, that is, top-level, legislatures (as opposed to local or
+          state-level legislatures). In the future we hope to include those too.
```
tmtmtmtm: 'hope to' is also strong — we _might_ do this, but we might not.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        </p>
+        <p>
+          Furthermore, we split the data for those down into
+          <strong>terms</strong>. Like legislatures themselves, the way this
+          works may vary from country to country. Terms have a start and end
+          date, and the data therefore describes the membership of the
+          legislation during that period. In most democracies, new terms begin
+          after every national election. Of course, changes still occur within
+          terms -- the point is, that’s how we break the data files down for
+          you.
```
tmtmtmtm: The last sentence here is quite confusing.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        <h2>
+          Politicians
+        </h2>
+        <p>
+          <strong>The richness of the data we have will depend on the source or
+          sources from which we’re getting it.</strong> We try to use official
+          sources, but this isn’t consistently useful across the world. Indeed,
+          the EveryPolitician project exists partly because in some places it’s
+          still depressingly hard to get even basic information about the
+          current legislature.
```
tmtmtmtm: &lt;nit&gt; the most recent isn't necessarily the same thing as the current…

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          legislation during that period. In most democracies, new terms begin
+          after every national election. Of course, changes still occur within
+          terms -- the point is, that’s how we break the data files down for
+          you.
+        </p>
+        <p>
+          This is pragmatic because we know online services and sites are
+          likely to be concerned with the most recent term -- that is, a
+          country’s <em>current</em> legislature.
+        </p>
```
tmtmtmtm: &lt;nit&gt; the most recent isn't necessarily the same thing as the current…

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        <h2>
+          Politicians
+        </h2>
+        <p>
+          <strong>The richness of the data we have will depend on the source or
+          sources from which we’re getting it.</strong> We try to use official
+          sources, but this isn’t consistently useful across the world. Indeed,
```
tmtmtmtm: The "we try to use official sources" part isn't really accurate. As an aspiration it might be, but there are lots of places where we use unofficial sources even if the official source might also be good.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          the EveryPolitician project exists partly because in some places it’s
+          still depressingly hard to get even basic information about the
+          current legislature.
+        </p>
+        <p>
+          But the absolute minimum data we need for a politician to be in
+          EveryPolitician is their <strong>name</strong> (by implication, we
+          also have their membership of the legislature and its term). Unless
+          the system they are in does not use such mechanisms, we’ll also have
```
tmtmtmtm: well, we _try_ to have. That doesn't mean we'll always know it.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          There may also be start and end dates if we know they did not serve
+          the full term.
+        </p>
+        <a name="about-names"></a>
+        <h3>
+          About names
+        </h3>
+        <p>
+          Names are a key part of any data set based around people. In the
+          basic CVS, the <code>name</code> will be the common name that we’ve
```
tmtmtmtm: we're on git these days ;)

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          <code>given_name</code>
+          <code>family_name</code>
+          <code>honorific_prefix</code>
+          <code>honorific_suffix</code>
+          <code>patronymic_name</code>
+          <code>sort_name</code>
+        </p>
+        <p>
+          However, in the CSV you’ll be getting these in whichever language was
+          available from the source of the data. Often this is the official
```
tmtmtmtm: I'm not quite sure what this is trying to say, but I don't think it's accurate.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          language of the country concerned, but that’s not always
+          straightforward because many countries have multiple official
+          languages. Furthermore, you can’t determine <em>which</em> language
+          it is from within the CSV — remember that CSV really is a coarse
+          version of the data — so instead you should use the JSON. In there
```
tmtmtmtm: There's a recurring undercurrent through of "I'm telling you what's in the CSV, but you really shouldn't be using it". We should either go the other way around, describing the JSON, but noting that a simplified version is available as a CSV, or soften these to a much more explicit "if you need X, then the JSON is probably better".

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          you’ll find that fields such as names are broken out by language code
+          (and often that will include transliteration in languages that are
+          not official languages of the country concerned, too).
```
tmtmtmtm: This part about official languages seems slightly odd or out of place.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          match them (or, as ever, use the JSON data instead).
+        </p>
+        <p>
+          In fact, we’ve even built an external service that consolidates all
+          the names into one CSV file (caution: that currently contains well
+          over 80,000 names and growing): see
+          <a href="https://github.com/everypolitician/everypolitician-names/">everypolitician-names</a>.
+        </p>
+        <h3>
+          IDs to other data sets, including WikiData
```
tmtmtmtm: WikiData shouldn't be camel-cased like this — it's just Wikidata (here and elsewhere)

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        <p>
+          Each identifier consists of an (<code>identifier</code>,
+          <code>scheme</code>) pair. For example, here’s an entry for the
+          identifier in WikiData (the database behind Wikipedia):
+        </p>
+        <pre>
+"identifiers": [
+  {
+    "identifier": "Q1091494",
+    "scheme": "wikidata"
```
tmtmtmtm: it's quite odd to give an example that isn't even a human here.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+  }
+]
+</pre>
+        <p>
+          (Incidentally, you can view WikiData entries like this:
+          <a href="https://www.wikidata.org/wiki/Q1091494">wikidata.org/wiki/Q1091494</a>,
+          or using <a href="https://tools.wmflabs.org/reasonator/?q=Q1091494">the Reasonator</a>.
+          This example is the Parliament of Åland.)
+        <p>
+          We do this for other entities (such as legislatures themselves) as
```
tmtmtmtm: perhaps worth noting that we're in the process of adding them for political parties/factions too?

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          well as politicians. Remember that external data sets might not be
+          complete, so it’s possible to have a useful identifier which isn’t
+          populated for <em>all</em> records. For example, WikiData only holds
+          data for entries that satisfy their “notability” criteria.
+        </p>
+        <p>
+          Some legislatures have adopted the good practice of assigning unique
+          IDs to their politicians, which can be useful when using their
+          official APIs. Whenever we can find IDs like this, we’ll add them as
+          identifiers in this way.
```
tmtmtmtm: again, I'd like this to have more of a participatory feel, asking people to let us know of sources for those.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        </p>
+        <h2>
+          Legislatures and terms
+        </h2>
+        <p>
+          For every country, we break the data down into
+          <strong>legislatures</strong> (for example, the “parliament” or
+          “congress” of the country). Currently, we’re only working with
+          national, that is, top-level, legislatures (as opposed to local or
+          state-level legislatures). In the future we might include those too.
```
tmtmtmtm: slight scoping confusion here. The 'those' refers to the thing in the brackets from the previous sentence, which tripped me up slightly.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+                  </li>
+                </ul>
+              </div>
+            </div>
+          </div>
+        </div>
+        <p>
+          <span class="superscript">&dagger;</span>
+          You can use fields marked &dagger; as IDs, although we can’t
+          guarantee they’ll always be unique <em>outside</em> this legislature.
```
tmtmtmtm: "Can't guarantee" may be a little too weak here. We can almost guarantee that they _won't_ be: lots are just things like `party_id: 1`. It's also a little odd that `id` itself isn't marked with a dagger (it's unclear whether it should be, or whether this description needs to be adjusted slightly)

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+          legislation during that period. In most democracies, new terms begin
+          after every national election.
+        </p>
+        <p>
+          This is pragmatic because we know online services and sites are
+          likely to be concerned with the country’s <em>current</em>
+          legislature, which means the data for politicians in the current
+          term. Often — but not always — this is also the data we’re most
+          likely to be able to source online, so if you see a country only has
+          one term in its data, it may well be the current one. We’re always
```
tmtmtmtm: I'd say "probably the most recent one" here.

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        <h2>
+          Politicians
+        </h2>
+        <p>
+          <strong>The richness of the data we have will depend on the source or
+          sources from which we’re getting it.</strong> We use a wide variety
+          of sources, some official and many more that are not. Indeed, the
+          EveryPolitician project exists partly because in many places it’s
+          still depressingly hard to get even basic information about the
+          current legislature.
```
tmtmtmtm: perhaps "about who the current legislators are"?

views/data_summary.erb
```diff
@@ -0,0 +1,355 @@
...
+        </p>
+        <a name="about-names"></a>
+        <h3>
+          About names
+        </h3>
+        <p>
+          Names are a key part of any data set based around people. In the
+          basic CSV, the <code>name</code> will be the common name that we’ve
+          got from the source, but we’re aware that names are not quite that
+          simple. We often have this name data:
```
tmtmtmtm: "Often" is perhaps a little strong. Perhaps something like "So we sometimes have separate fields for"?

views/data_summary.erb
tmtmtmtm: or even something like "Whenever we're made aware of" that still pushes the assumption that we're expecting people to contribute.

tmtmtmtm:
:+1:

PR #4254, opened 2016-02-29
WIP: Term cards with completeness box

tmtmtmtm:
Display term pages as cards of people, rather than table rows. 

For now we don't display the 'completeness box' at the top of the page — that will come next. 

Part of https://github.com/everypolitician/everypolitician/issues/289


tmtmtmtm:
* Example with lots of data: https://everypolitician-viewer-pr-4254.herokuapp.com/uk/commons/term-table/56.html
* Example with very little data: https://everypolitician-viewer-pr-4254.herokuapp.com/niue/assembly/term-table/15.html

tmtmtmtm:
@zarino I would have expected that the filtering would squish all the remaining cards together, but at the minute it leaves gaps (e.g. searching for Tallinn at https://everypolitician-viewer-pr-4254.herokuapp.com/estonia/riigikogu/term-table/13.html)

screen shot 2016-02-29 at 07 43 17

Is this what you were expecting to happen?

----

Also, we've lost the `start_date` / `end_date` display (compare Arto Aas, or Edgar Savisaar, here vs at http://everypolitician.org/estonia/riigikogu/term-table/13.html) — where do you think that should be displayed? (bearing in mind there could be more than one pair per person, most commonly if they change party, but occasionally if they change constituency — e.g. Alex Maskey at http://everypolitician.org/northern-ireland/assembly/term-table/4.html 

In theory this could also apply to name changes, (e.g. Pam Lewis → Pam Brown → Pam Cameron, also in the current NI Assembly), though we don't try to deal with that in the display at the minute. It might be nice to include those previous names, though, if only just so that searching / filtering works on them.

zarino:
Re: gappy layout while filtering – Yep, I knew this would happen. It's actually surprising hard to stop this happening and yet still have the items line up properly in rows. The answer is that I need to write some JavaScript to fix the gaps during each pass of the filter. I've added it to your list up top.

Re: dates – Is it just the stuff in the white part of each person's card that differs based on date? (ie: not the contact details in the pale green part?)

tmtmtmtm:
> The answer is that I need to write some JavaScript to fix the gaps during each pass of the filter. I've added it to your list up top

Ah. Surprisingly hard, indeed!

> Re: dates – Is it just the stuff in the white part of each person's card that differs based on date? (ie: not the contact details in the pale green part?)

Yes. I don't think we need to care about things like what someone's twitter handle _used to_ be. We should always just be showing the most up-to-date data for those elements.

zarino:
@tmtmtmtm Just so I understand, what's the use-case(s) for showing more than just a person's current party and constituency?

tmtmtmtm:
@zarino, as a Term page, it's primarily about all the memberships of that term — including changes. It's not about a snapshot at a single point in time: it's about what happened during that legislative period. 

This is particularly important in places where there's a separation between the legislature and the executive, and becoming a Minister means resigning your seat — so being able to see those people coming and going is very valuable (if this was only a snapshot of the current point, or at the end-date of a historic term, there are lots of people who wouldn't appear at all, either because they resigned mid-term, or because they were a mid-term substitute for someone who was a Minister for 18 months, for example). Or, similarly, in countries like Ukraine or Italy (or quite a few others) where there are regular mass changes of party/faction affiliations, it's very useful to be able to see those.

I suspect that part of the issue here is that although this is notionally a page about the Term, it's ending up having to de facto act as the source of information on the People too, leaving it unable to do either of these things as well as we might like.

I'm very open to other suggestions for different ways of splitting things up, adding other pages etc.

zarino:
@tmtmtmtm Do you think something simple like this would make the start and end dates clear enough?

screen shot 2016-03-03 at 15 27 48

I guess it could even extend to support people who moved from one party/location to another:

screen shot 2016-03-03 at 15 31 20

zarino:
@tmtmtmtm The branch now includes a flexbox-based grid layout (which solves the gappy layout bug when filtering – and falls back to a plain vertical list on old browsers), display of until/between/from dates for parties/areas, and lazy-loaded images.

Let me know if you spot any issues with any of those. I’ve tested the flexbox layout and the lazy loaded images in modern versions of Chrome, Firefox and Safari, with and without javascript support enabled.

As for the last 4 things on the list:

* **Passing real data to the stats display** – Can you / @chrismytton do this?
* **Avoiding display of "unknown" as a party** – Not sure of the best place to do this, maybe filtering it out in the `@csv.find_all` lines? Maybe another one for you / @chrismytton – the styling should just work if the person’s membership list is empty.
* **Lazy loading people** – I don't think we need to do this.
* **"More data" link on cards** – I'm not sure when/why this would be required?

tmtmtmtm:
@zarino — looking good, thanks! Wiring up the stats, and filtering out 'unknowns' is easy enough for @chrismytton or I to do. 

Will let you know if I spot any oddities.

zarino:
Also, just spotted, we should check whether all the things in the /images directory are still required – eg, the facebook / twitter images we added back when this branch was about adding icons to the table rows.

tmtmtmtm:
I've fixed up the tests, removed those left-behind images, and changed the area/group display to not show 'unknown' or empty values. 

So I think this is pretty much ready to go now, other than the completeness box, which I suspect we should do separately.

PR #4763, opened 2016-03-07
Present CSV and JSON download options on term page

zarino:
We use the Houdini CSS library from other mySociety projects to handle the dropdown behaviour without JavaScript. But because Houdini uses `id/for` attributes we also needed to tweak the `$.fixedThead()` module to deduplicate ids on cloned elements.

dropdown

Fixes everypolitician/everypolitician#324.

### DIFF
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -19,6 +19,21 @@
       });
     }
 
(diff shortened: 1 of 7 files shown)


tmtmtmtm:
:+1: Thanks @zarino!

PR #5258, opened 2016-03-15
Simpler wording for earlier/later links on term page

zarino:
Fixes everypolitician/everypolitician#293.

screen shot 2016-03-15 at 09 54 56


tmtmtmtm:
Yay! :+1:

Issue #5253, opened 2016-03-15
Toggle-able Person cards


tmtmtmtm:
We've gone through a couple of iterations of what to show on the 'Person' cards on a Term page. Currently they only show a couple of social-media links:

![screen shot 2016-03-15 at 08 13 25](https://cloud.githubusercontent.com/assets/57483/13771437/d0fdd464-ea85-11e5-837b-530e2d5e07eb.png)

Rather than trying to cram everything we know about them onto one card, these should be toggle-able in some way, to show some values for each of the data 'layers' we have for this legislature — e.g. biographical info (gender; birth/death dates); contact info (email; phone); social (facebook; twitter; linkedin; blog); profiles (i.e. external links _not_ owned by the person themselves — the Guardian; Journalisted; official parliament website etc); identifiers (IDs that don't map to profiles); etc.

These can then be toggled by clicking the relevant section of the 'completeness' bar at the top of the page (not currently live, or wired up — the design is [implemented but commented out](https://github.com/everypolitician/viewer-sinatra/blob/0ca3c9cd13386f229539eb6dbdb4c3fbd2b7d6d1/views/term_table.erb#L62-L104))

tmtmtmtm:
I suggest we do this by moving all this display logic out of the template, and passing through a processed data structure from `app.rb`, rather than simply a list of Person objects. As part of this we should also ensure that all those values are obtained via `EveryPolitician::Popolo` rather than extracting them manually from the JSON.

Initially we should probably still _read_ the term CSV file, and then filter the JSON based on the IDs of people in that, until we're completely confident that the process for generating the CSVs and for generating these people is identical.

zarino:
+1 for doing less logic in the template.

We'll need to investigate the best way of presenting the toggles for switching between the 5 data categories. (eg: toggle in one place that affects the whole page, duplicated toggles on each card that affect the whole page, or toggles on each card that affect only that card)

tmtmtmtm:
@zarino: Yep — I'm not sure yet what is likely to be best here. I think we can start by just having a global toggle from the completeness bar and then see if it's sensible/plausible to also toggle on a per-card basis. 

I have a slight fear that this is going to make filtering a little nasty too — does that filter against only what's currently _displayed_, or against everything on a card? And if the latter, will it be confusing for a card to be shown without it being obvious why (if the filter string is on a different view of it)?

zarino:
[...]

PR #5267, opened 2016-03-15
Use EveryPolitician::Popolo on the term page

chrismytton:
To clean up the templates a bit we want to switch to using [EveryPolitician::Popolo](https://github.com/everypolitician/everypolitician-popolo) and push more of the logic into the route handler.

Part of https://github.com/everypolitician/viewer-sinatra/issues/5253


app.rb
```diff
@@ -77,20 +78,44 @@
...
+    }
+  end
 
   @urls = {
     csv: csv_file.url,
     json: popolo_file.url,
   }
-  @data_sources = (popolo['meta']['sources'] || [popolo['meta']['source']]).map { |s| CGI.unescape(s) }
+
+  # TODO: Make this use EveryPolitician::Popolo once that supports the 'meta' field.
```
tmtmtmtm: as an in-between measure, should EveryPolitician::Popolo be able to give you 'raw' JSON (i.e. post-processed, as a data-structure, not just the underlying string)? That would at least avoid us having to parse the file multiple times?

chrismytton: Yep, ticketed here for now - https://github.com/everypolitician/everypolitician-popolo/issues/20

chrismytton:
This is now using `EveryPolitician::Popolo` for person information. There are now sections for "social", "bio", "contact details" and "identifiers" in the markup, but only "social" is being displayed currently. So this should be identical to the currently deployed version, though it will produce a big diff on viewer-sinatra because it will generate the extra markup for the other hidden sections.

Going to merge this then iterate on the controls for showing/hiding these sections in a separate pull request.

PR #5341, opened 2016-03-16
Toggle person-card sections

### LINKED ISSUE #5253: Toggle-able Person cards

zarino:
Fixes #5253.

You can now click the section labels under "What’s in this dataset?" to reveal the associated section in each person-card on the page below:

screen shot 2016-03-16 at 12 51 14

Also, when you search via the `.js-filter-input` above the cards, if your search term appears inside a hidden section of a person-card, that section will be temporarily unhidden (in addition to the already visible section) so you can more clearly see why the card matches your search term. Like in the following example, where a search for "q258" reveals the wikidata identifiers for the matching people, even though normally only the "social" section of each card would be visible:

screen shot 2016-03-16 at 12 58 43


chrismytton:
@MyfanwyNixon We could do with you working your magic on the text on this page. You can [view the page here](https://everypolitician-viewer-pr-5341.herokuapp.com/uk/commons/term-table/56.html). It currently says:

> Pick a subject to reveal the information we hold about each politician in the section below.

It's worth mentioning that (quoting @tmtmtmtm) "this is still only a subset (an often a tiny subset) of the information we hold about them".

MyfanwyNixon:
Woah, that's so cool. (That's not my copy, but it could be).

How about:
> <H2>What's in this dataset?</H2>

> As much as we know. Click one of the categories to get a taste of the information we have about the politicians in this dataset.

tmtmtmtm:
:+1: for "a taste of". Not sure we need the duplication of "in this dataset" though. Maybe worth making that something that conveys that we're always adding more, as well? "a taste of the information we have so far" or something like that?

MyfanwyNixon:
> #    What's in this dataset?

>   As much as we know. Click one of the categories to get a taste of the information we have so far.

Yes, I think it's ok to drop the mention of politicians as well, as you can see them quite clearly.

zarino:
I’ve made @MyfanwyNixon’s changes to the text.

chrismytton:
I've wired up the percentages now. For the time being I've just done it in the simplest way, i.e. counting how many people have *any* data for a given section.

screen shot 2016-03-17 at 10 24 43

tmtmtmtm:
The ordering of these is a little odd. I think it would be better to swap bio and social around.

tmtmtmtm:
I'm still not sure that the message is strong enough that we usually have more (often _much_ more) in the JSON. The "a taste of" is good, though a little subtle. Maybe we could add something below the darker section that explicitly points people at the JSON for these and others? In theory that message could vary based on whether we really do have more / how much more etc, so we don't necessarily need to find just one message for this.

MyfanwyNixon:
So, various scenarios:

Lots of data:
> Need more detail? There are plenty more fields in the JSON version of this dataset.

A bit more data:
> Hoping for something else? Check out the JSON version of this dataset.

No more, or very little more data:
> Prefer a different format? Check out the JSON version of this dataset.

chrismytton:
Biography section now comes first and is displayed by default.

screen shot 2016-03-17 at 17 06 49

tmtmtmtm:
There's a lot of things still to improve with this (particularly in terms of the code), but let's get it live and iterate.

PR #5349, opened 2016-03-16
Click a party name to show only people in that party

zarino:
Fixes everypolitician/everypolitician#340.

There is potential for a confusing situation when you filter by party and then start additionally filtering by text using the `js-filter-input`. (Your most recent action will take precedence, but the UI won't make that clear.)

@chrismytton: This appears to work fine for some countries (eg: the UK) but not others (eg: USA). I think maybe it's because I'm assuming the `mem.on_behalf_of_id` of a membership and the `mems.first[:group_id]` in the party list are the same, but they're not (eg: in the USA, the links up top have `data-party-filter="republican"` but the list items have `data-parties="party/republican"`) – why?

screen shot 2016-03-16 at 15 43 20

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -123,6 +123,7 @@ get '/:country/:house/term-table/:id.html' do |country, house, termid|
           end_date: mem.end_date
         }
         membership[:group] = org_lookup[mem.on_behalf_of_id].name if mem.on_behalf_of_id
+        membership[:group_id] = mem.on_behalf_of_id
         membership[:area] = area_lookup[mem.area_id].name if mem.area_id
         membership
       end,
--- a/public/javascript/ep-card-filter.js
+++ b/public/javascript/ep-card-filter.js
@@ -29,7 +29,7 @@ var CardFilter = function(){
   // When the UI is in its default position,
   // this is what the state should look like.
   var defaultState = {
-    facet: "biography",
+    facet: "bio",
     search: undefined,
     party: undefined
   }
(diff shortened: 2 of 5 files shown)


chrismytton:
> This appears to work fine for some countries (eg: the UK) but not others (eg: USA). I think maybe it's because I'm assuming the mem.on_behalf_of_id of a membership and the mems.first[:group_id] in the party list are the same, but they're not (eg: in the USA, the links up top have data-party-filter="republican" but the list items have data-parties="party/republican") – why?

Ah this is my bad, I said that the template only uses JSON now but actually the parties are still generated from the CSV. I'll pick this up and make the party list come from the CSV as well.

chrismytton:
OK I've changed things so that **everything** in the template really does come from the JSON files now. There is no longer anywhere in `app.rb` that uses the term CSVs. This should hopefully prevent any further confusion of this nature.

tmtmtmtm:
Once I've filtered by group, how do I get back to the full list again?

zarino:
@tmtmtmtm – does clicking a second time not clear the filter?

tmtmtmtm:
@zarino it does, but it was entirely unobvious that that's what I was meant to do…

zarino:
Agreed. The whole section is a bit clunky. But wary of spending more
time on it until we know how it gets used at all.

See also #5408.

tmtmtmtm:
Sure. How will we actually know how it's being used? Are we tracking the individual clicks, etc?

zarino:
Not yet! See #5420.

chrismytton:
@tmtmtmtm There doesn't seem to be a review app, but @zarino has updated this pull request so that it tracks which party you've clicked on and scrolls the member list into view after clicking the party.

tmtmtmtm:
When it scrolls down to the 'Members' section after selecting a group, I think it needs to do something to make it more obvious that this is just a subset of all members, and how to get back to the full set.

PR #5351, opened 2016-03-16
WIP Highlight people on homepage

### LINKED ISSUE #5263: Highlight some Person cards on the home page
We want to show off that we have a lot of data — ideally by example, not just by claiming it.

So pick some key People to highlight (possibly pseudo-randomly) on the home page and display the cards for them — letting you toggle around the different views of what types of data we have on them (per https://github.com/everypolitician/viewer-sinatra/issues/5253)

chrismytton:
:warning: Not ready to be merged yet!

This 3 featured person cards to the homepage. It needs @zarino to come in and make it pretty now.

This is what it looks like if I fiddle the CSS to show all the info on a person card:

screen shot 2016-03-16 at 16 56 11

Fixes #5263


zarino:
Currently looks like this:

screen shot 2016-03-16 at 16 28 14

@chrismytton – Note the incorrect page header – "Estonia"!?!?

@tmtmtmtm – I have a feeling you'll want to pick new wording for the subheading I've added. /cc @MyfanwyNixon

chrismytton:
@zarino I've rebased this onto master, so you'll need to to a hard reset if you want to make further changes:

```
git checkout 5263-highlight-people-on-homepage && git fetch origin && git reset --hard 5263-highlight-people-on-homepage
```

MyfanwyNixon:
I'd prefer:

The world's richest open dataset on politicians

tmtmtmtm:
Yay. Generally looking great. I think we might need to explain better what this section is though, or find a way to integrate it better with the page as a whole. Could probably do with something to make it obvious that the four featured sections are far from all we have.

We'll also want to change the legislative info section — now that it's being shown in a different context, it'll need to make sense there. Switching the constituency to the country would probably be a good first step. Though we probably something to show dates or similar as well would be useful — especially if we pick people with a long political history.  (Hmmm… actually this makes me realise that we probably want a 'History' tab RSN on the term page too, so that you can see which earlier / later terms the person was also part of…)

tmtmtmtm:
If we're going to make the claim of 'richest dataset', might also be nice to give an indication of scale somewhere. (We're currently at 2.6 million pieces of data — more if we include the 'positions' stuff that's still just in the 'unstable' section)

zarino:
Now looks like this:

screen shot 2016-03-17 at 09 26 39

MyfanwyNixon:
> ## The world's most extensive open dataset on politicians 
> Millions of pieces of data, including:

> [social media profiles] [biographical data] [Contact details] [Identifiers]

MyfanwyNixon:
Jinx!

zarino:
2.6 makes it sound much more scientific ;-)

MyfanwyNixon:
@zarino - no arguments from me, I was just future-proofing :)

tmtmtmtm:
If only we were capable of using technology to count them so the number could be automatically up to date at all times… ;)

chrismytton:
> If only we were capable of using technology to count them so the number could be automatically up to date at all times… ;)

Is that data being exposed anywhere that we can pull it from? Would certainly be good if that number was automatically kept up to date!

tmtmtmtm:
`bin/count_statements.rb` — though we should probably tweak that a little bit before using the number anywhere. At the minute it's counting any string value anywhere, which means it's counting things like 
```json
        {
          "lang": "bar",
          "name": "Aadu Must",
          "note": "multilingual"
        },
```  
as three statements.

tmtmtmtm:
I've added a `statement_count` to each entry in `countries.json`. We can use that to generate the "2.6 million pieces of data" statement (I think we should always round down to that form (2.6 million, 2.7 million etc), though it would be nice if the full count was available by hovering.

chrismytton:
I was hoping this would be mostly finished today but rawgit being down derailed me somewhat. Anyway, this is nearly there now, but still needs a bit of work so that the people on the homepage don't get all of the terms that they've been in listed as memberships. The people are read from `featured_people.yml` but that's currently empty, so nothing will appear on the homepage currently (I have left an example of what to put in that file though).

PR #5462, opened 2016-03-18
Switch to using GitHub raw urls directly

chrismytton:
RawGit has been down a couple of times lately, so instead of pointing to
that this changes things to point directly to GitHub. This does mean
that all JSON and CSV urls will have the content type text/plain (as
that's what GitHub serves them up with) but we can document that and
tell people to use RawGit if they need the correct content type.


chrismytton:
We haven't had anymore problems with RawGit recently, so I'm closing this for now. Should be easy enough to redo in the future if we do end up needing it.

PR #5625, opened 2016-03-21
Refactor person logic

chrismytton:
This was part of https://github.com/everypolitician/viewer-sinatra/pull/5351. I've broken it out into a separate pull request because the changes to the homepage in #5351 were getting hard to see between all the refactoring.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -11,6 +11,7 @@ require 'yajl/json_gem'
 require 'everypolitician/popolo'
 
 require_relative './lib/popolo_helper'
+require_relative './lib/people'
 
 Dotenv.load
 helpers Popolo::Helper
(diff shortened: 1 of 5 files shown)


tmtmtmtm:
@chrismytton I'm assuming most of this is still relevant? If so, I'll happily merge if you can rebase it.

tmtmtmtm:
Actually, having played about with this a bit, I'm not convinced this is the right abstraction here. 

We shouldn't really need the `People::Collection` class — a lot of this is working around shortcomings in `everypolitician-popolo`, and should really be solved there, rather than here.  Then we could have a much simpler View class to represent the Person, with the same sorts of methods you currently have for each of the toggle-able card views, but which shouldn't know anything about any collection it might be part of. The only slightly odd thing we need to resolve there is the identifiers, as that involves a calculation based on the collection, but I think that can be handled by passing the current `collection.top_identifiers` as an argument to `identifiers` instead.

tmtmtmtm:
superseded by major rewrite of entire system!

PR #5635, opened 2016-03-21
Use Popolo JSON exclusively for the term table

chrismytton:
Previously we were getting the person ids from the CSV and then mapping
those to the JSON. This changes that to instead get everything directly
from the popolo to avoid confusion where there are subtle differences
between the CSV and the JSON.

I tried to cherry-pick the appropriate commits but too much had changed to make it simple, so I just reimplemented on top of `master`.

Extracted from #5349

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -94,19 +94,15 @@ get '/:country/:house/term-table/:id.html' do |country, house, termid|
   area_lookup = Hash[popolo.areas.map { |a| [a.id, a] }]
   org_lookup  = Hash[popolo.organizations.map { |o| [o.id, o] }]
 
-  # Pull all unique parties out of memberships. We don't want to count
-  # memberships that finished before the end of the term. We then sort them by
-  # size and remap them to the format that the template consumes.
-  @parties = term_memberships.find_all { |mem| mem.end_date.to_s.empty? || mem.end_date == @term[:end_date] }
-             .group_by(&:on_behalf_of_id)
-             .map { |group_id, mems| [org_lookup[group_id], mems] }
-             .sort_by { |group, mems| [-mems.count, group.name] }
-             .map { |group, mems| { group_id: group.id.split('/').last, name: group.name, member_count: mems.count } }
-
-  # If we don't know the parties for anyone in a term then hide the parties section.
-  if @parties.length == 1 && @parties.first[:name] == 'unknown'
-    @parties = []
-  end
+  # Groups, ordered by size, with name and count of how many members
+  # they had at the end of the term. Don't include this section if
+  # all groups are unknown.
+  @group_data = term_memberships.find_all { |mem| mem.end_date.to_s.empty? || mem.end_date == @term[:end_date] }
+                .group_by(&:on_behalf_of_id)
+                .map { |group_id, mems| [org_lookup[group_id], mems] }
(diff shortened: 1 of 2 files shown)


app.rb
```diff
@@ -79,23 +78,33 @@
...
   popolo_file = EveryPolitician::GithubFile.new(@house[:popolo], last_sha)
   popolo = EveryPolitician::Popolo.parse(popolo_file.raw)
-  people = popolo.persons.find_all { |p| person_ids.include?(p.id) }.sort_by { |p| p.sort_name }
 
-  @parties = @csv.find_all { |r| r[:end_date].to_s.empty? || r[:end_date] == @term[:end_date] }
-                 .group_by { |r| r[:group] }
-                 .sort_by { |p, ms| [-ms.count, p] }
+  memberships = popolo.memberships.find_all { |m| m.legislative_period_id.split('/').last == id }
+
+  person_ids = memberships.map(&:person_id)
```
tmtmtmtm: Probably doesn't make a huge amount of difference as this variable is only used once, but as these things have a habit of expanding over time, and just to be clear what's going on, it might be nice to `.uniq` this, as a Person might have more than one Membership during a term.

tmtmtmtm: Also, as we're doing an `.include?` on this inside a loop immediately afterwards, probably good to make it a Set rather than an Array.

app.rb
```diff
@@ -79,23 +78,33 @@
...
-  memberships_by_person = popolo.memberships.find_all { |m| m.legislative_period_id == "term/#{id}" }.group_by(&:person_id)
+  memberships_by_person = memberships.group_by(&:person_id)
   areas_by_id = Hash[popolo.areas.map { |a| [a.id, a] }]
   orgs_by_id = Hash[popolo.organizations.map { |o| [o.id, o] }]
 
+  current_memberships = memberships.find_all { |mem| mem.end_date.to_s.empty? || mem.end_date == @term[:end_date] }
+                                   .group_by(&:on_behalf_of_id)
+                                   .map { |group_id, mems| [orgs_by_id[group_id], mems] }
+                                   .sort_by { |group, mems| [-mems.count, group.name] }
+
```
tmtmtmtm: `current_memberships` seems remarkably complex when all we're using it for now is creating `@parties`

app.rb
```diff
@@ -79,23 +78,33 @@
...
   popolo = EveryPolitician::Popolo.parse(popolo_file.raw)
-  people = popolo.persons.find_all { |p| person_ids.include?(p.id) }.sort_by { |p| p.sort_name }
 
-  @parties = @csv.find_all { |r| r[:end_date].to_s.empty? || r[:end_date] == @term[:end_date] }
-                 .group_by { |r| r[:group] }
-                 .sort_by { |p, ms| [-ms.count, p] }
+  memberships = popolo.memberships.find_all { |m| m.legislative_period_id.split('/').last == id }
+
+  person_ids = memberships.map(&:person_id)
+  people = popolo.persons.find_all { |p| person_ids.include?(p.id) }.sort_by(&:sort_name)
```
tmtmtmtm: I think I'd move this `sort_by` down to where we're generating the final `@people` — I don't think it matters at this point if they're sorted, and it makes this section a little harder to follow.

PR #5742, opened 2016-03-22
Make generating the person structure more concise

chrismytton:
This uses the explicit methods added in https://github.com/everypolitician/everypolitician-popolo/pull/22 to make creating the person structure for the term table template much more concise.


app.rb
```diff
@@ -149,40 +139,24 @@
       fb_username = person.facebook.split('/').last
       p[:social] << { type: 'Facebook', value: fb_username, link: "https://facebook.com/#{fb_username}" }
     end
-    if person.gender
-      p[:bio] << { type: 'Gender', value: person.gender }
+    p[:bio] << { type: 'Gender', value: person.gender } if person.gender
+    p[:bio] << { type: 'Born', value: person.birth_date } if person.birth_date
+    p[:bio] << { type: 'Died', value: person.death_date } if person.death_date
+    p[:contacts] << { type: 'Email', value: person.email, link: "mailto:#{person.email}" } if person.email
+    person.contact_details.each do |cd|
```
tmtmtmtm: The branch we're basing this off does already have `phone` and `fax` methods which we could request directly, rather than iterating over each contact asking "Are you a Phone?", "are you a Fax?": https://github.com/everypolitician/everypolitician-popolo/blob/24e5c5ebcdcc72517efc4ac6ff28fac24fcbd777/lib/everypolitician/popolo/person.rb#L31-L37

tmtmtmtm:
Much nicer! Other than slight simplification of phone/fax (and switching back to using master, now that that branch is merge), I think this is good to go.

Issue #5629, opened 2016-03-21
Don't show Groups section if there's only one


tmtmtmtm:
We [already](https://github.com/everypolitician/viewer-sinatra/pull/5345) hide this if everyone is 'Unknown', but we should also hide if it if everyone is 'Independent', or any other time there's only a single entry — e.g. http://everypolitician.org/alderney/states/term-table/2014.html

PR #5743, opened 2016-03-22
Hide 'Party Groupings' section if only one group

tmtmtmtm:
Don't show the list of Groups if there's only one entry (e.g. if
everyone is an Independent).

Closes https://github.com/everypolitician/viewer-sinatra/issues/5629


chrismytton:
:+1:

chrismytton:
I've merged this so it doesn't get lost in the long list of PRs!

PR #5744, opened 2016-03-22
Round down the number on the homepage

### LINKED ISSUE #5671: Round the homepage statement count down
We shouldn't be saying we have 2.7 million until we actually do: 

screen shot 2016-03-21 at 21 27 13

chrismytton:
Rather than saying we've got 2.7m statements when we've only got 2.68m
we instead say we've got 2.6m. This means we won't claim we've got 2.7m
statements until we actually do.

screen shot 2016-03-22 at 16 28 02

Closes #5671

### DIFF
--- a/lib/popolo_helper.rb
+++ b/lib/popolo_helper.rb
@@ -54,7 +54,8 @@ module Popolo
     end
 
     def number_to_millions(num)
-      (num.to_f / 100_000).floor.to_f / 10
+      result = (num.to_f / 100_000).floor.to_f / 10
+      result.modulo(1) < 0.1 ?  result.to_i : result
     end
   end
 end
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
Once we get to 3 million, I think it would be better if this said "3 million" rather than "3.0 million"

(You'll presumably also want to add the 'Closes …')

chrismytton:
@tmtmtmtm Yep good shout. I've change it so it displays as "3 million" rather than "3.0 million". I've also added some tests for the various different states.

chrismytton:
I've simplified the tests to just test as a plain module now rather than with `Rack::Test` as I originally did.

tmtmtmtm:
:+1:

PR #5745, opened 2016-03-22
Decode Facebook usernames before displaying on term page

chrismytton:
screen shot 2016-03-22 at 16 46 50

Because we're getting the username out of the url some usernames contain
UTF-8 characters that need to be decoded before they are displayed.

Fixes https://github.com/everypolitician/everypolitician/issues/358


tmtmtmtm:
I've a nagging feeling that this should probably live somewhere slightly different, but definitely Good Enough For Now™.

PR #6685, opened 2016-04-05
New comprehensive CardFilter class for showing/hiding cards on term page

### LINKED ISSUE #5407: Refactor term page filtering/segmenting/searching JavaScript into a single module
The following bits of the term page are currently interactive, but they’re handled completely separately:

* Clicking the social/bio/contacts/identifiers buttons in the green “What’s in this dataset?” bar, to flick through the various info sections on each person-card.
* Clicking the party/faction names at the top of the page, to filter the person-cards, leaving only cards of a particular party/faction.
* Typing free text into the search box, to filter the person-cards, leaving only cards containing the specified text.

Maybe it sort of makes sense that the first one is handled by its own little function (it's about altering the state of the contents of each card, rather than the visibility of the card as a whole).

But the other two functions are very similar, and since they’re currently handled completely separately, it’s possible to get into situations where they conflict (eg: filtering by party, and then entering a search term – the party name will still show as selected, even though the party filter has been entirely replaced by the free text search).

Maybe we should spend some time sorting this out. Especially if we ever plan on adding more interactivity to this page (eg: to filter/show/hide cards based on another criteria).

zarino:
Fixes #5407 by combining the previously separate `js-filter-input` and `[data-active-section]` handling into a single module.

State is saved to the URL hash, and new history entries are created as you filter, so you can bookmark or share pages as you see them, and can navigate backward and forward in history with your browser’s back/forward buttons.

### DIFF
--- a/public/javascript/ep-card-filter.js
+++ b/public/javascript/ep-card-filter.js
@@ -9,24 +9,106 @@
     });
     return found;
   }
+
+  // Inverse of $.param()
+  // http://stackoverflow.com/a/26849194/3096375
+  $.deparam = function(str) {
+    return str.split('&').reduce(function(params, param) {
+      var paramSplit = param.split('=').map(function (value) {
+        return decodeURIComponent(value.replace('+', ' '));
+      });
+      params[paramSplit[0]] = paramSplit[1];
+      return params;
+    }, {});
+  }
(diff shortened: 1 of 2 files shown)


chrismytton:
Looks good to me! :+1:

PR #9582, opened 2016-05-20
Show biography section on person cards by default

chrismytton:
This wasn't working correctly due to a typo: `biography` instead of `bio` as the default section name.

screen shot 2016-05-20 at 16 52 11

Fixes https://github.com/everypolitician/everypolitician/issues/400


tmtmtmtm:
👍

PR #10620, opened 2016-06-02
tweaked so column widths sum to < 100%

### LINKED ISSUE #10619: two-column layout always renders as single column
Tucked in the lovely `_layouts.scss` are `first-column` and `second-column`, but the widths sum to just over 100% so they always render as a single column.

We're not using it currently, but I am about to :-)

davewhiteland:
closes #10619 

For @zarino to sanity-check please

### DIFF
--- a/views/sass/_layouts.scss
+++ b/views/sass/_layouts.scss
@@ -7,14 +7,14 @@
     }
     .first-column {
         @media (min-width: $medium_screen) {
-            width: 48.5%;
+            width: 47.5%;
             margin-right: 3%;
         }
 
     }
     .second-column {
         @media (min-width: $medium_screen) {
-            width: 48.5%;
+            width: 47.5%;
         }
     }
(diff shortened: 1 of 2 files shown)


zarino:
@davewhiteland: This looks fine to me. Really, the standard way of doing gutters and columns these days is to have a negative margin on the parent, and then an equal padding on both sides of the column (resulting in a double gutter between the columns). But as a quick fix for what's already there, this is fine.

davewhiteland:
@zarino OK thanks. For now I'll put this in since it's functionally what I want, and we (you) can revisit when the page I'm using it for comes up for proper design work.

zarino:
Yep. It's likely we’ll introduce a set of proper grid classes when we implement https://github.com/everypolitician/everypolitician/issues/388, since the standard mysociety footer a variety of column widths.

Issue #10523, opened 2016-06-01
Move the 'download' information to a separate page


tmtmtmtm:
We currently offer 'Download' links from two different pages (the country page, and an individual term page). They've evolved separately from each other, and are confusing in several key ways. To make them more understandable we need to add more text to them, but this is already unwieldy, and we certainly don't have room to add _more_ options (e.g. the names file, or other types of downloads; explaining the difference between the current *term* (potentially with multiple memberships for people) and the current *members*, etc.).

So, instead, let's add a separate Downloads page for each country, listing and explaining all the different options, and either linking to or transcluding part of the docs explaining things like rawgit, fetching by sha, keeping up to date via webhooks etc.

davewhiteland:
This now implemented:
First in #10637 which introduced a bug (in the sinatra &rarr; wget &rarr; gh-pages process), fixed in quick succession by two slightly ugly PRs: #11632 and #11636

PR #10637, opened 2016-06-02
Issues/10523 download page

davewhiteland:
Addressing #10523 -- intercepting download buttons and sending to a country-specific download page instead. Needs design loveliness but this is a proof-of-concept to work from.

Not yet done (suggestions please): the existing "download links" are _still_ buttons which seems tricksy, because they now go to a page which contains download buttons.

Also I'm not sure about diving deep into a download page with many terms on it for term data, because the JSON alternative (to the term's CSV) will be above the fold.


davewhiteland:
Following screenshots drawing attention to main points: 

First this is a whole-screenshot (reduced font) of an "nice" example: Ireland, because it's got a single house with just two terms in it (so the CSV-per-term makes sense; in countries where we've only got one term the talk of terms is potentially confusing).

Basically this is the "download" page for a country, and you can see there are *five* download buttons which do indeed link to distinct rawgit-hosted CSV or JSON files. This _feels_ broadly right to me because we do have five distinct downloads; `countries.json` is of course common to all the country pages, but this seems like a reasonable place for it to turn up.

Note also I'm describing the term pages as summaries, as in "View Summary" buttons -- otherwise these could be links (as they are on the country's own page). That catch there is a user who's come from such a page (having clicked "Download data" to get here) then perhaps expects clicking on "32 Dáil" to download that data, instead of being pinged back to the page they just came from. Hence not presenting the names as links at all, and offering a "view summary" button instead.

whole-download-page

davewhiteland:
Here's the hijack going on (i.e., the purpose of this pull request). The "Download" button used to link directly to a data file; now it goes to the download page (see previous comment for an example). This happens in two places: on the country file (which used to link to the JSON, i.e., per-legislature, which is how the country page is divied), and on each term page (where it could link to JSON or CSV).

The advantage of the interstitial download page is that the JSON/CSV distinction is not being offered to the user without full what's-available context (compare with currently, when looking at the country page, you're shown "download JSON" without necessarily even knowing the per-term CSVs are available).

I've casually styled these buttons differently, but -- for now -- they're still buttons and they have the download icon on them. I _think_ that's OK; they're a downloady-call-to-action even if they are not actually initiating the download on that click. Um, right?

download-link

download-link-terms

davewhiteland:
And here's a problem. The way I've implemented this, if you click on a term page "download" link, I'm sending you to the country's download page (good!), but specifically to the CSV for **the term you were on**. If, like this case (Germany) there are many terms, the other interesting things on this page are _above the fold_.  The red arrow is pointing more or less at the top of the #anchor (OK actually it's`id` now because modern) element (a list item, which seems appropriate). The JSON alternative (and indeed `names.csv` too) is above the fold, scrolled off to place the "cursor" on this position.

Might need design magic to address: possibly the left column floats-stickily to a `fixed` position while it's on-screen, but that might be a bit ugh.

download-page-under-fold

tmtmtmtm:
@davewhiteland don't get too bogged down trying to do that sort of magic. Just get a page that shows all we want to show. People are perfectly capable of scrolling to the part the want. 

As well as simple pragmatism, this is also important because the more we presume what they want and try to be fancy, the more we'll get it wrong — e.g. if they clicked on the Download button on a term page that had been filtered in some way.

davewhiteland:
Indeed; it feels like something we might need to fix one day if we discover people bounce when coming from the term pages on countries with many terms but not other routes, etc. And until then, it's fairly clear you've landed in the midst of a page.

davewhiteland:
And here's the other decision I built into this: the download page iterates over houses, so for bicameral systems you get the preamble/explanatory text twice. I can live with this; it's slightly, publicly inelegant but it's also not distracting unless the country has very few terms, when they repetition might be present on the single screen. Here's the example of Australia, scrolled down and font-shrunk to show what's happening (see **Just politicians' names** in this long screenshot).

screen shot 2016-06-03 at 16 02 39

tmtmtmtm:
> until then, it's fairly clear you've landed in the midst of a page.

I was suggesting that they _don't_ land somewhere in the middle — they just come in at the top, and work out what they want after that.

tmtmtmtm:
With suitable handwaving, I'd be tempted to collapse that long term list in some way (perhaps only if there's more than N entries?)

Seems like it could get unwieldy fast, particularly for somewhere with lots of terms, across two houses (for the US we'll conceivably have over 200 entries there once we pull in more historic data…)

davewhiteland:
Yes, I like!

Although that's where coming in from a known term might be handy -- to collapse the list to a max of say three entries (with clear "show all") so that that one is displayed (default to latest being the most desired). 

Otherwise from some term pages we'd be linking to a page to download data that isn't _displayed_ on that page on arrival, which seems rude.

Fairly straightforward with JS: want me to add that now?

tmtmtmtm:
Yes — if they came from a term page, defaulting to that one being visible or highlighted or the like is a lot better than just scrolling the page to that place.

wfdd:
There'd still be a lot of wasted space when expanded.  I think it'd make more sense to lay the name and whole legislature data sections side by side and the terms underneath those two, with the term list spanning multiple columns, like on country pages.

davewhiteland:
Now with that layout (switched to smaller buttons on the terms since this is less piggy layout). I don't doubt that we could use some designer love on this once it's done. (weeny font for layout demo only)

screen shot 2016-06-03 at 16 51 43

davewhiteland:
Ah, forgot to add screenshot here: this is now the page you get if you click on "Download data" on the  7th Bundestag term. Things to note: no longer going to an "anchor" tag within the page (user can scroll), only some (3, currently) of many terms are shown (the rest are collapsed), and the "wanted" one, if it exists, is highlighted for extra helps. Needs design love, but this is proof-of-concept/working demo (e.g. https://everypolitician-viewe-pr-10637.herokuapp.com/germany/download?house=bundestag&term=7 )

screen shot 2016-06-08 at 11 38 50

tmtmtmtm:
@davewhiteland Is design time needed before this can go live, or something we can layer on later?

davewhiteland:
@tmtmtmtm depends entirely on how horrible this is to look at. I think it's humane enough to go live, with a ticket for designers to fix it up -- those two buttons (which I've put in a `button-group` for seeing the summary or downloading the data of each term looks a bit bodgey, and I suspect a designer would want to border or shade the elements themselves in some way, because they run into each other a bit raggedly at the moment).

But I'd say the benefit of having this interstitial download page at all is greater than the pain of it not being very beautiful.

Someone with a little more patience than me might fix the two places where I didn't use jQuery's expected functions and used a less elegant solution: see these self-confessed room-for-improvements (didn't seem worth wasting more time fighting them than I did so far):

https://github.com/everypolitician/viewer-sinatra/pull/10637/commits/8a0b57a87a0572af5463cd3070d8d832d0fb75da#diff-5818829e680f939d37c7f05010659745R165

https://github.com/everypolitician/viewer-sinatra/pull/10637/commits/8a0b57a87a0572af5463cd3070d8d832d0fb75da#diff-5818829e680f939d37c7f05010659745R190

Both `.index()` and `.splice()` seemed obvious fits, both failed to work the way I expected after reading the docs twice and trying variations on the invocations, and both I could code round just to get it working :-(

tmtmtmtm:
Main issue I have with it is that the h3 of the legislature name isn't very distinctive, so it's quite difficult to see what's going on in a bicameral country. If there's a quick way to make that stand out more, I'm happy for this to go live as is, and be tidied up more later. (Have you already requested design time for this?)

zarino:
@davewhiteland – This looks good enough to ship, for sure. If you really want me to bikeshed something, I guess I would make the green "Show all terms" button *much* less prominent. Right now it looks like the primary CTA on the page, and, if I didn't pay attention, I'd click it expecting it to initiate a download. "Show all 18 terms" can just be a plain (orange) link, positioned *after* the line up of the default 3 terms.

Also not clear what the "View Summary" buttons would do, or why I'd want to click them. My first instinct would be to drop those "View summary" buttons, and instead just make the term titles (eg: "7th Bundestag") a link instead. (You can wrap an `<a>` around the `<h3>`, and the layout should continue to work fine.)

davewhiteland:
@zarino thanks; the reason I did the summaries as a button and not as a link off the term title, is that I anticipate a common route to this page is _from_the (term) page (having clicked on a "download data" button that you're proposing I link to. So when the user clicks on the term they want here, are they expecting to get data or to be sent back to the page from which they came? That's why I unlinked the title and used buttons instead to make that a forced what-do-you-want decision. Thoughts?

davewhiteland:
Still needs some designer love.

wfdd:
The query strings are tripping up the static site generator - try downloading the data of any country from everypolitician.org.

davewhiteland:
@wfdd indeed, thanks: Fixed in quick succession by two slightly ugly PRs: #11632 and #11636

zarino:
> Still needs some designer love.

Looks fine to me!

PR #11461, opened 2016-06-13
Issues/437 redirect docs pages

davewhiteland:
This replaces the existing (bad, old) docs pages within http://everypolitician.org with redirections to the corresponding pages in their new home on http://docs.everypolitician.org.

This is everypolitician/viewer-static#7  (wrongly attempted on the static files) re-done on this repo (viewer-sinatra), which is effectively the generator.

`wget` seems to play nicely here, in that (as expected) it does not parse/obey the `meta refresh` redirection, so the pages stick.

Closes everypolitician/everypolitician#437

Sanity check: there are nine doc pages being redirected.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -22,16 +22,14 @@ end
 wjson = File.read('world.json')
 WORLD = JSON.parse(wjson, symbolize_names: true)
 
+DOCS_URL = 'http://docs.everypolitician.org'
+
 # Can't do server-side redirection on a GitHub Pages-hosted static site, so the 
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
I'm not convinced that `redirect_page` should have logic for working out where to redirect to (i.e. the remapping of `about.html`. Wouldn't this better as an entirely agnostic function which also takes an argument of where to redirect to? 

I'd probably also explicitly name that `soft_redirect` or something similar, to make it clearer that it's a client-side redirect, and is distinct from sinatra's inbuilt redirection methods.

PR #11632, opened 2016-06-16
remove query params from the download links

### LINKED ISSUE #11631: remove query param from download page
the query params are being captured by wget, drop them for now because this has broken all downloads

davewhiteland:
can reintroduce them when I've sorted out not having wget record them as part
of the page name
(highlighting won't work on select terms for now)
closes #11631


davewhiteland:
considering this an urgent bugfix so merging now

Issue #11915, opened 2016-06-21
Name-of-parliament should appear on the term page(s)


davewhiteland:
e.g.: Hungary's Országgyűlés:

http://everypolitician.org/hungary/ &larr; "Országgyűlés" here...

http://everypolitician.org/hungary/assembly/term-table/40.html &larr; ...but not here

If we've got the name, we should be using it both for SEO and confirmation for user that this is the right place. Note that sometimes the name has been included in the term title, in which case it's not an issue, e.g., http://everypolitician.org/ukraine/verkhovna-rada/term-table/8.html

I originally ticketted this in the wrong repo: everypolitician/everypolitician#409

octopusinvitro:
This is how it looks like in the main title:

![possibility](https://cloud.githubusercontent.com/assets/2157089/16737792/e20fbb8e-478a-11e6-9fe2-397791133f8d.png)

tmtmtmtm:
How does this look for somewhere with long names, like St Vincent and the Grenadines?

davewhiteland:
Maybe join it with a line break (I wonder if our designers would be horrified by that idea)  -- we're already not being precious about the vertical space on these pages

octopusinvitro:
What about inside an h3?

![possibility](https://cloud.githubusercontent.com/assets/2157089/16738406/6d7746ae-478d-11e6-92f3-6039e14d84dc.png)

davewhiteland:
Yup, works for me 👍

octopusinvitro:
ok! Gonna make the PR now :)

wfdd:
An `h3` isn't more semantic here.  Headings demarcate [sections of content](https://html.spec.whatwg.org/multipage/semantics.html#headings-and-sections), which the top bar isn't; neither the country name nor the name of the house should be contained in a heading.  Also, an `h2` without a parent `h1` is logically unsound, if not semantically invalid.

wfdd:
I do like the presentation though :)

PR #12941, opened 2016-07-11
Add house name in the term table under country name

octopusinvitro:
If there is no country there will be no house name either, so I added a conditional under the title.
Wrapped it with an h3 tag.
It works for long names in country or house name.

Closes: #11915


octopusinvitro:
This is how it looks:

possibility

tmtmtmtm:
👍

PR #12987, opened 2016-07-12
add anchor tag to URL and send house+term as query vars into download page

### LINKED ISSUE #11911: show the selected term data on the download page
When going to the download page for a legislature from a specific term, should show (and highlight) the selected term where there are multiple terms. Currently we only show the three most current terms (the rest are hidden, and available on a "show all" button click; but the point is if there's a selected term that one should always be amongst the ones shown automatically. 

Note: this was implemented using query variables but I subsequently urgently pulled the query variables out of the URL in everypolitician/viewer-sinatra#11631 because they were fatally breaking the (viewer-sintra &rarr; wget &rarr; viewer-static creation of) the website.

So really this issue is:

* pass the selected term information into the downloads page, and consume that

It would nice to have this functionality back; maybe by passing the link as the #-target (which arguably makes sense already for the house in bicameral legislatures, so you go to the right part of the download page), or else by using query params but changing the way `wget` captures them.

This was implemented with this code which at time of writing is still in place on the download pages: https://github.com/everypolitician/viewer-sinatra/blob/658ce7aac57e3f6ef6b6f904469edf312dec42a6/views/country_download.erb#L126-L197

I originally ticketted this in the wrong repo: everypolitician/everypolitician#445

davewhiteland:
Anchor tag added for jumping to the right legislature.

Intercept clicks with a specific term to target using Javascript, so it works fine but not when we're running wget.

Closes #11911


wfdd:
> Intercept clicks with a specific term to target using Javascript, so it works fine but not when we're running wget.

`wget` doesn't execute JS, if `wget` using the parametrised URL is what you were trying to avoid.  You could do this on load and not have to worry about race conditions (which you might get on old smartphones, like mine...).

davewhiteland:
Yes, the JS interception is because the only reason for passing the term+house parameters are so that JS on the subsequent page can make client-side changes: specifically if that page has many terms, it hides all but three (and makes sure the "current" term is one of those (ideally in the middle)).

And yes, we don't want `wget` to pick up these query variables, which is why it's being done in JS this way.

And yes I agree the JS should be moved into a document-ready section, will do.

davewhiteland:
@tmtmtmtm @chrismytton invitation to review please (some squashing necessary afterwards perhaps, but showing my working)

public/javascript/main.js
```diff
@@ -376,5 +376,21 @@ $(function(){
 
           $(this).data('chart', chart);
       }
-  })
+  });
+
+  $("a.download-with-term").on("click", function(e){
```
zarino: Such a minor thing, but not sure you need a the `a` in `$("a.download-with-term")`

public/javascript/main.js
```diff
@@ -376,5 +376,21 @@ $(function(){
...
+    if (term && house) {
+        // inject query vars into URL before the target anchor
+        e.preventDefault();
+        location = $(e.target).attr('href')
+          .replace('#', "?" + $.param({term: term, house: house}) + "#");
+    }
+  });
+
+  if (typeof page_init_function === 'function') {
+      page_init_function();
```
zarino: The javascript naming convention everywhere else in this project is camel case. Is there a reason to use underscores here instead?

views/sass/_custom.scss
```diff
@@ -431,3 +431,7 @@ table.definition {
   cursor: default;
 }
 
+.download-terms__highlighted {
+  background-color: #fffadd;
```
zarino: Any chance you could make this colour a Sass variable, in `_variables.scss`. Maybe something like `$colour_pale_yellow`?

public/javascript/main.js
zarino: Also, if this a class that’s *only* used to hook JS behaviour onto, it’s good practice to prepend it with `js-`, so that it’s more obvious what the class is for, when people are refactoring code. So that’d make the selector look like: `$('.js-download-with-term')`.

zarino:
Looks good to me. As discussed in Slack, having a download page for each legislature probably makes most sense in the long run, for a number of reasons. I've created a new ticket for that at https://github.com/everypolitician/viewer-sinatra/issues/13599. 

But this is fine for making the existing download page a little better in the meantime.

tmtmtmtm:
@davewhiteland if it's good enough for Zarino, it's good enough for me :)  :shipit: 

I know it's not new to this PR, but I'm a little scared by that whole `max_visible_terms` logic. That all seems like it should be _lot_ simpler (or at least more encapsulated somewhere). What was the story with that?

davewhiteland:
@tmtmtmtm The `$button_for_reveal` parameter is ugly, but feels like more work than is needed right now to generalise away from this being used on for terms on the download page. See what you think. Certainly nicer than it was.

tmtmtmtm:
@zarino this has changed a bit since you last looked at it. Does it still all look OK to you?

public/javascript/main.js
```diff
@@ -376,5 +376,62 @@ $(function(){
...
+
+// collapse_displayed_items:
+// If there are too many items being displayed in the list, hide
+// all but a few, making sure the target element (if there is one)
+// is included in those that are not hidden. Add a button for
+// revealing the hidden items.
+// Note this is used on the download page, where (currently) there
+// may be more than one such list (e.g., one for each legislature).
+
+function collapse_displayed_items(
```
zarino: @davewhiteland Sorry to be *that guy*, but all of the other variables and function names in viewer-sinatra use camelCase. Why doesn't this function (and the variables inside it)?

public/javascript/main.js
```diff
@@ -376,5 +376,62 @@ $(function(){
...
+    $list_of_elements,   // list of elements that need to be collapsed
+    $target_element,     // item within that which needs to be highlighted
+                         // ...if none (which is OK) take the first
+    hidden_class_name,   // CSS class used to distinguish collapsed elements
+    min_threshold_items, // only collapse if more elements than this
+    max_display_items,   // number of elements to show when collapsed
+    $button_for_reveal   // button that is added to reveal
+  ) {
+  if ($list_of_elements.size() > min_threshold_items) {
```
zarino: Use `.length` instead of `.size()`<sup>[1]</sup>

<sup>[1]</sup> https://api.jquery.com/size/ – “The .size() method is deprecated as of jQuery 1.8. Use the .length property instead.”

views/country_download.erb
```diff
@@ -132,68 +132,42 @@ var max_visible_terms = 6;
...
+  // highlight the specific term (if there is one)
+  // collapse terms if there are many
 
-// if there are more than `max_visible_terms` terms, hide them:
-// display `reduced_visible_terms`
-//   * selected term in the middle if possible
-//   * otherwise the most recent
+  $(".country__legislature").each(function(){
+    var house_slug = $(this).data('house');
+    var $term_list_items = $('#terms-' + house_slug + ' li');
```
zarino: Small thing, but you call them `items` here, but `elements` on the next line. Would make more sense to me to standardise. eg: `$term_list_items` and `$wanted_list_item`

views/country_download.erb
```diff
@@ -132,68 +132,42 @@ var max_visible_terms = 6;
...
+      $wanted_term_element, // note: may be undefined
+      'hidden-term',
+      max_visible_terms,
+      reduced_visible_terms,
+      $button_for_reveal
+    );
+  });
+
+  $('.hidden-term').hide();
+  $('.button.toggle-term-display').on("click", function(e){
```
zarino: Don't need `.button` here.

views/country_download.erb
```diff
@@ -132,68 +132,42 @@ var max_visible_terms = 6;
...
+    e.preventDefault();
+    var house = $(e.target).data("house");
+    $('#terms-' + house + " li.hidden-term").slideToggle();
```
zarino: Don’t need `li` here.

views/country_download.erb
```diff
@@ -132,68 +132,42 @@ var max_visible_terms = 6;
...
+      max_visible_terms,
+      reduced_visible_terms,
+      $button_for_reveal
+    );
+  });
+
+  $('.hidden-term').hide();
+  $('.button.toggle-term-display').on("click", function(e){
+    e.preventDefault();
+    var house = $(e.target).data("house");
```
zarino: I think you want `$(this)` here, instead of `$(e.target)`.

In most situations, they’re equivalent. But imagine the situation where we wrap a `<span>` around the text content in the `toggle-term-display` element – clicking the `span` will trigger the `click` listener on the parent button, but `$(e.target)` will still be set to the `span`, which you don’t want. `$(this)` is what you want, as it will always be whatever element the event listener was bound to (`.toggle-term-display` in this case).

public/javascript/main.js
```diff
@@ -376,5 +376,62 @@ $(function(){
...
+// collapse_displayed_items:
+// If there are too many items being displayed in the list, hide
+// all but a few, making sure the target element (if there is one)
+// is included in those that are not hidden. Add a button for
+// revealing the hidden items.
+// Note this is used on the download page, where (currently) there
+// may be more than one such list (e.g., one for each legislature).
+
+function collapse_displayed_items(
+    $list_of_elements,   // list of elements that need to be collapsed
```
zarino: `$list_of_elements` is slightly misleading here. I initially assumed it was a jQuery object containing an actual list element (like a `<ul>`). In reality, the "list" is the jQuery object itself (which isn't actually a list, in any meaningful sense of the word).

Perhaps rename to `$elements` to avoid confusion?

Or, `$items`, since you use the term `items` more frequently inside this function, and the fact that the variable name starts with a `$` already implies that it’s a jQuery object containing one or more DOM elements.

views/country_download.erb
```diff
@@ -132,68 +132,42 @@ var max_visible_terms = 6;
...
+      $term_list_items,
+      $wanted_term_element, // note: may be undefined
+      'hidden-term',
+      max_visible_terms,
+      reduced_visible_terms,
+      $button_for_reveal
+    );
+  });
+
+  $('.hidden-term').hide();
```
zarino: Seems to me like this line, and the `.toggle-term-display` click listener after it should be moved inside `collapse_displayed_items()` since that's where all the rest of the DOM elements are manipulated.

zarino:
@davewhiteland – I've left a few (!!) comments – sorry, once I started, I figured I might as well finish. Feel bad making you fix them all – would you like me to do them?

public/javascript/main.js
davewhiteland: @zarino as I'm working through making these changes, I just want to comment on this one because it's the only one I disagree with: I think you're a bit cruel to say that the result of `$('#terms-blah li')` isn't a list in any meaningful sense of the word; I think it is, because the "jQuery object" is effectively (more or less their words) "an ordered set of elements"... which is pretty much the sense of the word "list" no? The sense of it here is that the function accepts such a list, one of which is the target to be highlighted. 

But... needing to revert to such pedantry suggests that it makes much more sense for this function to stop being so general, *admit* it's always going to be a jQuery object containing the parent `ul` element, and take it from there. Then all my talk of `items` where I'm _pretending_ they might not be `li` elements goes away. So good spot after all ;-)

zarino: 👍

davewhiteland:
@zarino or @tmtmtmtm this is just front-end JS, now superlovelier than at the start. Thanks for the refinement comments, now much tidier and tighter: please merge if cool with it all.

Note the Travis failures are Rubocop complaining about things in. e.g., Sinatra's gems, which seems _a bit harsh_ regards this PR so I believe you can ignore them.

Preview on https://everypolitician-viewe-pr-12987.herokuapp.com/ e.g.

Specific layout/highlighting behaviour of the term pages:

* https://everypolitician-viewe-pr-12987.herokuapp.com/united-states-of-america/download.html behaves with no extra magic: collapses terms, but no term highlighted

* https://everypolitician-viewe-pr-12987.herokuapp.com/united-states-of-america/download.html?term=101&house=house#legislature-house 101st Congress highlights right term, button operation is correctly independent of the other button on the Senate's terms later in the page

* https://everypolitician-viewe-pr-12987.herokuapp.com/north-korea/download.html?term=13&house=national-assembly#legislature-national-assembly works fine with a single term

* https://everypolitician-viewe-pr-12987.herokuapp.com/cyprus/download.html?term=10&house=house-of-representatives#legislature-house-of-representatives works fine with only two terms

* https://everypolitician-viewe-pr-12987.herokuapp.com/iceland/download.html?term=2007&house=assembly#legislature-assembly six or fewer terms is not too many, so terms aren't collapsed, but highlighting still works

tmtmtmtm:
@davewhiteland if you rebase this over master (which you'll need to do anyway, as there are conflicts), the travis errors should hopefully disappear…

I think you probably also want to add some lovely screenshots here, rather than just linking to a staging site that will vanish after merge.

views/country_download.erb
```diff
@@ -123,77 +123,41 @@
...
   var match = location.search.match(new RegExp("[?&]" + key + "=([^&]+)(&|$)"));
   return match && decodeURIComponent(match[1].replace(/\+/g, " "));
 }
 
-var max_visible_terms = 6;
-var reduced_visible_terms = 3; // best odd, so wanted term is in middle?
-var wanted_house = get_q_param('house');
-var wanted_term = get_q_param('term');
+var maxVisibleTerms = 6;
+var reducedVisibleTerms = 3; // best odd, so wanted term is in middle?
```
tmtmtmtm: Why the hesitant '?' at the end of this?

views/country_download.erb
```diff
@@ -123,77 +123,41 @@
...
+var wantedHouse = getQueryParam('house');
+var wantedTerm = getQueryParam('term');
+var pageInitFunction = function(){
 
-// for each house...
+  // for each house...
+  // highlight the specific term (if there is one)
+  // collapse terms if there are many
```
tmtmtmtm: Now that the code has been factored out to elsewhere, you've ended up with a comment that simply repeats in English what the code below it is doing, which is almost always a bad smell. If the code still isn't sufficiently legible to make the comment obsolete, then there's still more refactoring to do. Or, if it is, you should drop the comment.

PR #13334, opened 2016-07-19
WIP track usage of interactive page elements

### LINKED ISSUE #5420: Track usage of interactive page elements
We should use google analytics events to track how many people are using the interactive elements on the term page in particular.

Eg: do people spot the download button? Do people use the filter box? Do people use the person-card section toggle? Do people click party names at the top to filter by party? Do people navigate between terms using the Earlier/Later links, or the dropdown menu?

ondenman:
Added tracking to interactive elements on term_table page: prev/next term links, term dropdown list, filter field, download button, section-toggles. 

Tracked elements are given the `data-ga-track-click` attribute.

Example: `<button data-ga-track-click='Download button'>…`

JS finds all elements with this attribute and attaches an event handler.

Closes #5420


public/javascript/main.js
```diff
@@ -377,4 +377,16 @@ $(function(){
...
+
+  // Event tracking
+  // Clicks on elements with data-ga-track-click attribute will be recorded
+  $('[data-ga-track-click]').on('click', function(){ 
+    trackClick( $(this).attr('data-ga-track-click')); 
+
+    function trackClick(eventDescription){
+      eventDescription = eventDescription || 'Undescribed event.';
+      ga('send', 'event', 'user interaction', eventDescription, document.title);
+    }
```
chrismytton: Seems odd to define this function *in* the event handler and rely on JavaScript's function hoisting. I think it would be better to move this function out of the event handler.

views/term_table.erb
```diff
@@ -246,3 +246,5 @@
             here's <a href="http://docs.everypolitician.org/contribute.html">how to get that fixed</a>.</p>
         </div>
       </div>
+      
+
```
chrismytton: Trailing whitespace klaxon! 🚨 

Probably best to remove this so the diff only contains the changes you've made.

tmtmtmtm:
@ondenman You should explain more in the PR description here _what_ you've actually done (rather than _how_ you've done it). It's not clear from the description, for example, whether this actually adds tracking to anything, or only the ability to track things in future.

views/term_table.erb
```diff
@@ -12,14 +12,14 @@
...
-                <a href="<%= term_table_url(@country, @house, @prev_term) %>" class="term-navigation__prev">
+                <a href="<%= term_table_url(@country, @house, @prev_term) %>" class="term-navigation__prev" data-ga-track-click="Previous term link">
                     <i class="fa fa-arrow-left"></i>
                     Earlier
                 </a>
               <% end %>
 
               <span class="term-navigation__jump">Jump to:
-                  <select class="js-navigation-menu">
+                  <select class="js-navigation-menu" data-ga-track-click="Navigation menu">
```
zarino: Interesting. Probably want to bind to the `change` event of a `<select>`, rather than `click`.

views/term_table.erb
```diff
@@ -70,25 +70,26 @@
             <div class="container">
                 <div class="grid-list grid-list--vertically-center grid-list--narrow-columns">
                     <div class="data-completeness">
-                        <span class="section-toggle section-toggle--selected" data-section-toggle="bio">
+                        <span class="section-toggle section-toggle--selected" data-section-toggle="bio" 
+                        ="Selection toggle: bio">
```
zarino: Errant line break and equals sign here?

zarino: Or, rather, a missing `data-ga-track-click`…

views/term_table.erb
```diff
@@ -112,12 +113,12 @@
 
                 <span class="person-card-filter">
                     <label for="filter-input" class="fa fa-search"></label>
-                    <input type="text" class="js-filter-input" id="filter-input" placeholder="Search by name, place, etc…">
```
zarino: Removing this `id` attribute will mean that clicking on the preceding `<label>` won't focus the input any more. I assume you didn't mean to remove the `id` attribute?

public/javascript/main.js
```diff
@@ -377,4 +377,16 @@ $(function(){
           $(this).data('chart', chart);
       }
   })
+
+  // Event tracking
+  // Clicks on elements with data-ga-track-click attribute will be recorded
+  function trackClick(eventDescription){
+    eventDescription = eventDescription || 'Undescribed event.';
+    ga('send', 'event', 'user interaction', eventDescription, document.title);
```
zarino: Worth getting input from @zarino @wrightmartin @nick on what `eventCategory`, `eventAction`, `eventLabel`, and `eventValue` we should send here, with an eye to keeping it flexible enough for future types of event tracking, but also simple enough to draw useful conclusions via the Analytics reporting UI. [(Docs here)](https://developers.google.com/analytics/devguides/collection/analyticsjs/events)

ondenman: The JS isn't going to work at the moment as some of the interactions navigate away from the page. [Knowing when the hit has been sent](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#transport)

zarino: Probably best to force the Analytics call to be synchronous (ie: give it a `hitCallback` parameter) and then perform the default browser action (eg: navigating to the next page) only once we know the Analytics call has either completed or failed.

https://github.com/mysociety/fca-investment-tool/blob/master/web/public/js/analytics.js (mysoc team login required) is a pretty bulletproof wrapper we made previously, that uses jQuery deferred objects to keep the callback chaining super obvious.

ondenman: I think you referenced the wrong nick upthread.

@jacksonj04 Any thoughts on what to put in the event fields? At the moment I have:

  `hitType: 'event',`
  `eventCategory: ‘user interaction',`
  `eventAction: eventDescription,`
  `eventLabel: document.title`

ondenman: Removed tracking from events that lead off page. Will implement later today or tomorrow.

jacksonj04: @ondenman Those event parameters look as good a starting place as any, and more or less in line with what we've done elsewhere.

zarino: Yes, sorry, I got the wrong Nick!

ondenman:
Current implementation does not use deferred objects when tracking js-navigation-menu. Instead, it fires a ga() when the element is placed in focus. (Element has event already bound to it on line 281. Could replace anonymous function on line 281 with named function. Would then be able to redefine the function within the tracking code or using element.off(). Have experimented and it looks like .off() can only be used with named functions as arguments.)

public/javascript/main.js
```diff
@@ -377,4 +376,81 @@ $(function(){
...
-});
+
+// Google Analytics event tracking
+$('[data-ga-track-select]').on('focus',function(event){
+    ga('send', 'event', 'focus', $(this).attr('data-ga-tack-select'), document.title);
+})
+
+$('[data-ga-track-click]').on('click', function(event){
+    event.preventDefault();
+    event.stopPropagation();
```
zarino: Out of interest – why `stopPropagation()`?

ondenman: Yes, I should remove that. It's left over from my attempts at trying to prevent the dropdown list from firing.

views/term_table.erb
```diff
@@ -112,12 +112,12 @@
...
+                    <input type="text" class="js-filter-input" id="filter-input" placeholder="Search by name, place, etc…" data-ga-track-click="Filter field">
```
zarino: Should probably be tracking the `input` event here, rather than `click`. Firstly because you can tab to an input without clicking. And secondly because it's the act of actually *changing* the input’s value that we’re interested in tracking.

public/javascript/main.js
```diff
@@ -377,4 +374,99 @@ $(function(){
...
+    typingTimer = setTimeout(doneTyping, doneTypingInterval);
+});
+
+//on keydown, clear the countdown 
+$input.on('keydown', function () {
+  clearTimeout(typingTimer);
 });
+
+function doneTyping () {
+  ga('send', 'event', 'user input', $(this).attr('data-ga-tack-change'), document.title);
```
zarino: `data-ga-track-change` not `data-ga-tack-change`

views/term_table.erb
```diff
@@ -112,12 +112,12 @@
 
                 <span class="person-card-filter">
                     <label for="filter-input" class="fa fa-search"></label>
-                    <input type="text" class="js-filter-input" id="filter-input" placeholder="Search by name, place, etc…">
+                    <input type="text" class="js-filter-input" id="filter-input" placeholder="Search by name, place, etc…" data-ga-track-change="Filter field">
```
zarino: Maybe "filter term politician grid" or something might be a more useful label for the event?

(Just thinking, when we're looking at the reports in Google Analytics, it'll be much more obvious what and where the "filter term politicians" event came from, than the vague sounding "filter field")

views/term_table.erb
```diff
@@ -112,12 +112,12 @@
...
                 </span>
 
                 <div class="download-options">
                   <!-- download link as button? -->
-                  <a class="button button--quarternary" href="../../download.html">
+                  <a class="button button--quarternary" href="../../download.html" data-ga-track-click="Selection toggle: bio">
```
zarino: Should be `Download term data` rather than `Selection toggle: bio`

views/term_table.erb
```diff
@@ -70,25 +70,25 @@
             <div class="container">
                 <div class="grid-list grid-list--vertically-center grid-list--narrow-columns">
                     <div class="data-completeness">
-                        <span class="section-toggle section-toggle--selected" data-section-toggle="bio">
+                        <span class="section-toggle section-toggle--selected" data-section-toggle="bio" data-ga-track-click="Selection toggle: bio">
```
zarino: Wherever you’ve used `Selection` in the `data-section-toggle` attributes here, I think you mean `section`, since the toggles toggle "sections" of the card, not "selections".

views/term_table.erb
```diff
@@ -12,14 +12,14 @@
...
-                <a href="<%= term_table_url(@country, @house, @prev_term) %>" class="term-navigation__prev">
+                <a href="<%= term_table_url(@country, @house, @prev_term) %>" class="term-navigation__prev" data-ga-track-click="Selection toggle: bio">
                     <i class="fa fa-arrow-left"></i>
                     Earlier
                 </a>
               <% end %>
 
               <span class="term-navigation__jump">Jump to:
-                  <select class="js-navigation-menu">
+                  <select class="js-navigation-menu" data-ga-track-select="Selection toggle: bio">
```
zarino: Should be `Jump to term by date` or something, rather than `Selection toggle: bio`

views/term_table.erb
```diff
@@ -12,14 +12,14 @@
         <div class="container">
             <div class="term-navigation">
               <% if @prev_term %>
-                <a href="<%= term_table_url(@country, @house, @prev_term) %>" class="term-navigation__prev">
+                <a href="<%= term_table_url(@country, @house, @prev_term) %>" class="term-navigation__prev" data-ga-track-click="Selection toggle: bio">
```
zarino: Should be `Show previous term` or something, rather than `Selection toggle: bio`

views/term_table.erb
```diff
@@ -32,7 +32,7 @@
               </span>
 
               <% if @next_term %>
-                <a href="<%= term_table_url(@country, @house, @next_term) %>" class="term-navigation__next">
+                <a href="<%= term_table_url(@country, @house, @next_term) %>" class="term-navigation__next" data-ga-track-click="Selection toggle: bio">
```
zarino: Should be `Show next term` or something, rather than `Selection toggle: bio`

public/javascript/main.js
```diff
@@ -279,9 +278,7 @@ $(function(){
 
   $('.js-sortable').sortable();
 
-  $('.js-navigation-menu').on('change', function(){
-    window.location.href = $(this).val();
-  });
+  $('.js-navigation-menu').on('change', deferredTracking);
```
zarino: You've removed `window.location.href = $(this).val();` from this `change` handler, but not introduced anything to replace it inside `deferredTracking`.

`deferredTracking` probably needs to either work out that it's been called on a `<select>` rather than an `<a>`, or you need to make `deferredTracking` return the deferred promise, so you can attach your own completion callback up here.

public/javascript/main.js
```diff
@@ -377,4 +374,99 @@ $(function(){
...
+function deferredTracking(event){
+    event.preventDefault();
+    var href = $(this).attr('href') || false;
+
+    eventDescription = $(this).attr('data-ga-track-click') || 'Undescribed event';
+
+    var deferred = analytics.trackEvent({
+        hitType: 'event',
+        eventCategory: event.type,
+        eventAction: eventDescription,
```
zarino: Just want to check you’re sure about the order of `eventCategory` and `eventAction` here?

As it stands, we'll be recording an `eventCategory` like `"click"` and an `eventAction` like `"Selection toggle: identifiers"`.

Typically you'd use them the other way round – you’d use `eventAction` to record the action the user took (`"click"`) and `eventCategory` to record the type of object they were interacting with (`"Selection toggle: identifiers"`).

zarino:
@ondenman I’ve left a few more comments. Let me know if you want to go over them in a hangout.

ondenman:
Quicker and easier to start again rather than resolve conflicts... history is a mess too!

Issue #13297, opened 2016-07-18
Work out how to display executive positions when known


tmtmtmtm:
# Problem

In many countries we now have in the source data information about who holds key positions: so for the UK we know who is Prime Minister, Chancellor of the Exchequer, Home Secretary etc. We don't currently show that information anywhere. We should.

# Proposal

We have a few possible options. 

1. We simply add this information to a person's card on the Term pages — either as a new 'view' of that, or in the header. 

2. We 'hoist' the key people (e.g. Cabinet-level members, in countries where that is applicable) to the top of the list.

3. We add a whole new set of pages for this.

My suspicion that that although we want to evolve towards #3 over time there's a quick win from a combination of #1 and #2 whilst we get there (which probably requires page-per-person first)

# Acceptance Criteria

This ticket is currently only for working out what adding these could/should look like. Implementation can be another ticket. So we're really just looking for a design that will enable people to see at a glance all the people listed on http://www.parliament.uk/mps-lords-and-offices/government-and-opposition1/her-majestys-government/ when looking at http://everypolitician.org/uk/commons/term-table/56.html

# Not Required

In an ideal world, this will _also_ make it obvious that, say, David Cameron _was_ the Prime Minister during the same term (etc), but that isn't necessarily necessary for a first version. (We'll want to find a way of displaying things like that fairly quickly though, so it would be good if this version wasn't a dead end that made it much harder to get to there).

This will only need to work for countries where the Cabinet members are also part of the legislature. Handling countries where there's a separation of powers that precludes that can come later.

# Prerequisites

n/a

# Related Issues

n/a

# Due Date

n/a

# Design?

Yes — this is _entirely_ about design at this stage.

# Text?

Ideally what's going on will be so self-evident that we don't need any explanatory copy, but let's see :)

# Bloggable?

Once implemented across multiple countries this is probably worth shouting about.

zarino:
### Notes from conversation with Tony:

* This is about displaying a politician’s membership of the *executive branch* of the country (rather than—or in addition to—the *legislative branch*)
* We don't have this info for many countries yet, but will get more soon.
* The UK is a bit weird in that you can be *both* an MP for a constituency (a legislative branch position) *and* a member of the Cabinet (an executive branch position).
* In most countries, you *can’t* be a member of the executive *and* a member of the legislature at the same time. (This is why Barrack Obama doesn't appear on [EveryPolitician USA](http://everypolitician.org/united-states-of-america/)).
[...]

PR #13598, opened 2016-07-25
Tabs to switch between term pages for legislatures/houses in a single country

zarino:
Adds tabs to the top of the term_table pages, serving the dual purpose of making it clear which legislature/house is currently selected, and also making it easier to switch between legislatures/houses without having to go back "up" to the country-level page.

**Country with more than one legislature:**

screen shot 2016-07-25 at 14 15 11

screen shot 2016-07-25 at 15 09 39

**Country with only one legislature:**

screen shot 2016-07-25 at 15 08 09

screen shot 2016-07-25 at 15 09 14

Part of #13297, but actually quite beneficial to deploy on its own, since it improves the current experience.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -70,8 +70,5 @@ DEPENDENCIES
   sinatra!
   yajl-ruby
 
-RUBY VERSION
-   ruby 2.0.0p648
-
 BUNDLED WITH
    1.12.5
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -1,3 +1,18 @@
+<div class="page-section page-section--green page-section--no-padding">
+    <div class="container text-center">
+        <ul class="house-tabs">
+          <% @country[:legislatures].each do |house| %>
+            <li>
+                <% latest_term = house[:legislative_periods].sort_by{ |term| term[:start_date] }.last %>
+                <a class="house-tab <% if @house == house %>house-tab--active<% end %>" href="<%= term_table_url(@country, house, latest_term) %>">
+                    <%= house[:name] %>
+                </a>
+            </li>
+          <% end %>
+        </ul>
+    </div>
+</div>
+
 <div class="page-section" id="term">
     <div class="container text-center">
         <h1><span class="avatar"><i class="fa fa-university"></i></span> <%= @term[:name] %></h1>
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -3,8 +3,7 @@
         <ul class="house-tabs">
           <% @country[:legislatures].each do |house| %>
             <li>
-                <% latest_term = house[:legislative_periods].sort_by{ |term| term[:start_date] }.last %>
-                <a class="house-tab <% if @house == house %>house-tab--active<% end %>" href="<%= term_table_url(@country, house, latest_term) %>">
+                <a class="house-tab <% if @house == house %>house-tab--active<% end %>" href="<%= term_table_url(@country, house, house[:legislative_periods].first) %>">
                     <%= house[:name] %>
                 </a>
             </li>
--- a/views/layout.erb
+++ b/views/layout.erb
@@ -61,7 +61,6 @@
                         <a href="/"><span class="header-logo">EveryPolitician</span></a>
                       <% end %>
                     </h2>
-                    <% if @house %><h3><%= @house[:name] %></h3><% end %>
                 </div>
                 <div class="site-header__navigation site-header__navigation--primary">
                     <nav role="navigation">
--- a/views/sass/_components.scss
+++ b/views/sass/_components.scss
@@ -513,6 +513,17 @@ p.lead {
     @extend .inline-list;
     padding-top: 1em;
 
+    // Make the tabs scroll horizontally on small screens.
+    // TODO: Make this not suck when the currently selected tab is
+    // hidden off to the far right of the line.
+    @media (max-width: $small_screen_max) {
+        text-align: left;
+        white-space: nowrap;
+        overflow: auto;
+        margin: 0 -1em;
+        padding: 1em 0.5em 0 0.5em;
+    }
+
     @media (min-width: $medium_screen) and (min-height: 600px) {
         padding-top: 1.5em;
     }
@@ -537,4 +548,8 @@ p.lead {
 .house-tab--active:hover,
 .house-tab--active:focus {
     background: #fff;
+}
+
+.source-credits a {
+    @extend .break-word;
 }
--- a/views/sass/_components.scss
+++ b/views/sass/_components.scss
@@ -498,12 +498,43 @@ p.lead {
 
 // Little round social icons for membership tables
 .person-link {
-  display: inline-block;
-  width: 24px;
-  padding-top: 24px;
-  height: 0;
-  overflow: hidden;
-  vertical-align: middle;
-  border-radius: 100%;
-  margin-right: 0.3em;
+    display: inline-block;
+    width: 24px;
+    padding-top: 24px;
+    height: 0;
+    overflow: hidden;
+    vertical-align: middle;
+    border-radius: 100%;
+    margin-right: 0.3em;
+}
+
+.house-tabs {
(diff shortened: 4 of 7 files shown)


Gemfile.lock
```diff
@@ -70,8 +70,5 @@ DEPENDENCIES
   sinatra!
   yajl-ruby
 
-RUBY VERSION
```
zarino: I think this change was added because I ran `bundle install` before starting work. Not sure whether it's needed or not? @tmtmtmtm @chrismytton

tmtmtmtm:
Thanks @zarino. I'm having difficulties getting this running at the minute, as `nokogiri` is complaining about `libxml2` stuff, so whilst I'm wrestling with that, it would be helpful if you could also add a screenshot for a country where we only have a single legislature, and, as you've added some changes for narrow screens, a couple of parallel screenshots for those.

views/term_table.erb
```diff
@@ -1,3 +1,18 @@
+<div class="page-section page-section--green page-section--no-padding">
+    <div class="container text-center">
+        <ul class="house-tabs">
+          <% @country[:legislatures].each do |house| %>
+            <li>
+                <% latest_term = house[:legislative_periods].sort_by{ |term| term[:start_date] }.last %>
```
tmtmtmtm: These are already pre-sorted (though in most-recent-first order), so you just need a `house[:legislative_periods].first` there (which also means you can probably inline it rather than setting up a `latest_term`).

tmtmtmtm:
@zarino The behaviour of always jumping to the most recent term for a country surprised me a little here — e.g. when I was on `/australia/representatives/term-table/41.html` and didn't go to the same term in the Senate.

Was it an explicit decision that that's what the behaviour _should_ be?

zarino:
> @zarino The behaviour of always jumping to the most recent term for a country surprised me a little here

@tmtmtmtm: Yeah, it's not really what I wanted it to do. In an ideal world it would switch to the term, I guess, with the most overlap with the current term you're viewing. (I'm assuming terms don't necessarily increment at the same times, or have synchronised IDs between legislatures.)

tmtmtmtm:
Yeah — in some countries they'll pretty much always be in sync, but in others they'll be quite significantly different. Plus of course, they're not always two current houses of the same legislature.

Most overlap seems like an interesting way to go about it…

zarino:
> Most overlap seems like an interesting way to go about it…

Yeah, at least then you end up staying at roughly the same period in time. A simpler alternative might be to switch to the term with a start date closest to the start date of the term you're looking at.

tmtmtmtm:
> A simpler alternative might be to switch to the term with a start date closest to the start date of the term you're looking at

I think I'd be more inclined to make that a check on the _end_ date rather than the start date, unless there's a case I'm overlooking? (I'm imagining a scenario where the Upper house has 6 year terms, and the lower house 3 year terms, where if I'm looking at the current term in the Senate, and then switch to the lower house, I'd end up on the previous term, rather than the current one…)

tmtmtmtm:
I'm also not sure whether we should just go live with what we currently have, and potentially surprise some users in the meantime whilst we fix this up, or whether this is worth spending a bit of extra time doing the right thing now…

zarino:
> I'm also not sure whether we should just go live with what we currently have […]

It's hard to tell when we don't have any analytics data yet on how people navigate these pages.

I'm typically in favour of small incremental changes. But I'm a bit worried this is a sufficiently annoying thing that, if it happens, you'll just think the site is broken. Or, more perniciously, you might get switched to the wrong term without noticing.

Whatever we pick, there's always going to be a chance of people landing on the wrong term. Your suggestion of going by the end date is interesting, and certainly makes sense when you're switching from the *most recent* term. I just wonder whether it also makes sense when you're looking at historical terms?

tmtmtmtm:
> I'm typically in favour of small incremental changes. But I'm a bit worried this is a sufficiently annoying thing that, if it happens, you'll just think the site is broken. Or, more perniciously, you might get switched to the wrong term without noticing.

Yep — that's my fear too. If a change clearly adds value with no negatives, I'm all for pushing it through, even if that value is miniscule. But here I'm worried that it adds possible negative value as well as the positive, so I'm more cautious. 

> Your suggestion of going by the end date is interesting, and certainly makes sense when you're switching from the most recent term. I just wonder whether it also makes sense when you're looking at historical terms?

I _think_ it does. The one case where it might be odd is if you're looking at the first term in a country and then switch house, and end up on the second term of the other house. But that one doesn't worry as much as the opposite.

What this might all be telling us, of course, is that these tabs shouldn't be taking us to term pages _at all_, but just to the front page for that legislature. And the problem is that there's currently no such thing…

zarino:
> What this might all be telling us, of course, is that these tabs shouldn't be taking us to term pages at all, but just to the front page for that legislature.

Two worries about this:

1. The UI convention of tabs is that they switch the content below the tabs. Even though, technically, the tabs in this PR are navigating you to a new page, visually, the top half of the new page looks just like the top half of the old page, so the effect marries up with how you'd expect tabs to work. If the tabs lead you to a completely different looking page, the illusion will be broken. So they probably shouldn't be styled like tabs.
2. If the tabs take you to a "choose a term" page, then we're forcing *everyone* to make two clicks to get to the term they want (first click on the tab, then click on the term you want). Whereas, if clicking the tab at least *attempts* to send you to the term you're most likely to want, we could hope that *at least some people* would get to their desired destination in one click, and at worst, the remainder would get there in two clicks.

tmtmtmtm:
I think we're uncovering something quite interesting here, as now I'm unsure whether these _should_ be tabs or not. 

To me the primary thing we're dealign with here is the ability to switch to another sub-entity of the current country, and only secondarily to switch to an equivalent page — i.e. I currently see this as being something that's much more connected to what's _above_ it, than to what's _below_ it.

I suspect part of the difficulty in reasoning about this is because we mostly only really _have_ one type of page per legislature, and we currently mostly only one type of 'entity' we're dealing with. 

And whilst I definitely think we should work with what we currently have, rather than what we're hoping to have in the future, I'm a little way of introducing patterns that will completely break in other contexts if we can spot up-front something that would more naturally work both now and in the future.

To spell this all out a little more (as much for my own sake, as anything else), the only "Organization" (for lack of more precise terminology) we're dealing with here is a Legislature at the minute. In countries where we currently have more than one entity, those are of two types:

1. bicameral legislatures (e.g. [Australia](http://everypolitician.org/australia/), where the two are essentially sibling houses, which may or may not have their terms in sync (e.g. Australia where we do, or XXX where we don't).  [Note also that we may have much more historic data for one house than the other, such as with [Cameroon](http://everypolitician.org/cameroon/)]
2. historic legislatures (e.g [British Virgin Islands](http://everypolitician.org/british-virgin-islands/)

Soon we will have a third, connected but more distinct, version with executive level ones, e.g. the UK Cabinet. 

At a future data we may also have sub-national legislatures etc too (e.g. the legislatures of each of the US States). [NB: I don't think we need to pay a lot of attention to this yet, as that's probably quite a way off, and we'll likely have to rethink a lot of other core stuff when we get to that stage. Listing here purely for completeness]


Within each of those we currently only really have the Term-table pages. Reasonably soon it's highly likely that we'll have other types of things too. In practice we _do_ currently have other pages too, except we're careful not expose those on the static site. Interestingly the main one of those is now broken: https://everypolitician-viewe-pr-13598.herokuapp.com/australia/senate/wikidata 

The breakage seems to be from an earlier change, which we'll need to track down, but the interesting question is whether _that_ page should also have a way to get to the other house. My initial impression is that it should, but I guess I do expect that that would take me to the equivalent Wikidata page for the House of Representatives, rather than a 'front page' for it, which somewhat contradicts where I started this comment from, so I think I need to go away think about this a lot more!

tmtmtmtm:
I've fixed the broken Wikidata page in https://github.com/everypolitician/viewer-sinatra/pull/13743; might be worth rebasing this over that so we can reason better about what that page should do under this PR…

PR #14077, opened 2016-08-05
Thank browserstack in readme

zarino:
Browserstack gives us free cross-browser / cross-device testing, in exchange for an acknowledgement in the README of projects for which we use it.

I just noticed, the EveryPolitician.org README is missing the acknowledgement.


tmtmtmtm:
Thanks @zarino.

Issue #10276, opened 2016-05-29
Can't ctrl/cmd-A in search box


wfdd:
I'm assuming there's a two-way event listener that updates the search box from the address bar, so the instant I press cmd-A, keyup is detected, the new search term is pushed to the address bar and immediately pulled back into the search box, resetting my cursor.

tmtmtmtm:
@zarino, can you spot a quick fix for this?

zarino:
@wfdd is right – we’re binding an listener to the `keyup` event in that search box, which then overwrites the text input with the current search term after a short delay:

https://github.com/everypolitician/viewer-sinatra/blob/d0aaaa40f708483c9f374d2ef98e67a562e4658e/public/javascript/ep-card-filter.js#L186-L190

I think the simplest fix here is to only change the text input value if it *needs* to be changed. (ie: add another condition to [`this.setSearch`](https://github.com/everypolitician/viewer-sinatra/blob/d0aaaa40f708483c9f374d2ef98e67a562e4658e/public/javascript/ep-card-filter.js#L147-L157) that bails out early if `newSearch === newSearch`).

zarino:
Note to self: I wonder whether we could bind to the `input` event here instead.

PR #14341, opened 2016-08-10
More obvious card filtering UI on term table pages

### LINKED ISSUE #5408: Attach term page filters/buttons/search to the top of the window when scrolling
We currently let people filter/show/hide person-cards by clicking party names near the top of the page, by clicking the “What’s in this dataset” buttons in the green section above the person-cards, and by entering free text into the search box immediately above the person-cards.

Depending on the number of parties listed, this essentially means that the controls for manipulating the visibility of the cards are spread over an 800–1100px tall area, and—worse—when you scroll down the page to interrogate the cards in more detail, all of the filters become completely inaccessible.

Perhaps, when you scroll, the filters could attach themselves (in a more screen-height-efficient layout) to the top of the window – like the term table header used to when we had politicians displayed in tabular format on the old EP.org.

<del>This might be easier to do once #5407 has brought the logic for all of the filters into one place, conceptually.</del> – Done!

### LINKED ISSUE #6273: It's not obvious that you can click the dataset sections
The four items in the "What’s in this dataset?" section on a term page (e.g. http://everypolitician.org/finland/eduskunta/term-table/37.html) aren't obviously clickable.

screen shot 2016-03-31 at 13 54 04

zarino:
Fixes #6273 by introducing a new dropdown menu that you click to switch the visible "section" of each person-card (eg: bio, contacts, social).

Fixes #5408 by attaching the new filter bar to the top of the window when scrolling.

There's space to add a "Group:" dropdown (ie: a reworking of #5349) but I've left that to be done in a different PR.

Screenshots don't really do the scrolling awesomeness justice, but we can try:

### Desktop width, no filters active

screen shot 2016-08-10 at 14 44 09

### Desktop width, filtered by search term

screen shot 2016-08-10 at 14 50 44

### Narrow device, filtered by search term, and also scrolled down to show how the filters attach to the top of the screen

(Sorry, arrows missing on the dropdown because Chrome device imitation is rubbish – they're there on the real devices)

screen shot 2016-08-10 at 14 46 06

All tested and working in all the major browsers. Degrades nicely in IE8.

### DIFF
--- a/public/javascript/ep-card-filter.js
+++ b/public/javascript/ep-card-filter.js
@@ -99,10 +99,16 @@ var CardFilter = function(){
     $('[data-active-section]').attr('data-active-section', state.facet);
     $('.js-show-facet').val(state.facet);
 
-    if( typeof state.search === 'undefined' ){
+    if( state.facet === defaultState.facet ){
+      $('.js-show-facet').removeClass('js-filter-trigger--changed');
+    } else {
+      $('.js-show-facet').addClass('js-filter-trigger--changed');
+    }
+
+    if( state.search === defaultState.search ){
       $('.js-filter-target--hidden').removeClass('js-filter-target--hidden');
       $('.js-person-card__section--visible').removeClass('js-person-card__section--visible');
-      $('.js-filter-input').val('');
+      $('.js-filter-input').val('').removeClass('js-filter-trigger--changed');
 
     } else {
       $('.js-filter-target').each(function(){
@@ -126,7 +132,7 @@ var CardFilter = function(){
         });
       });
 
-      $('.js-filter-input').val(state.search);
+      $('.js-filter-input').val(state.search).addClass('js-filter-trigger--changed');
     }
 
     // Other scripts might want to do something special once they know
--- a/public/javascript/ep-card-filter.js
+++ b/public/javascript/ep-card-filter.js
@@ -97,8 +97,7 @@ var CardFilter = function(){
   this._updateUI = function(){
 
     $('[data-active-section]').attr('data-active-section', state.facet);
-    $('[data-section-toggle]').removeClass('section-toggle--selected');
-    $('[data-section-toggle="' + state.facet + '"]').addClass('section-toggle--selected');
+    $('.js-show-facet').val(state.facet);
 
     if( typeof state.search === 'undefined' ){
       $('.js-filter-target--hidden').removeClass('js-filter-target--hidden');
@@ -194,8 +193,8 @@ $(document).ready(function(){
       window.cards.saveState();
     });
 
-    $('[data-section-toggle]').on('click', function(){
-      var section = $(this).attr('data-section-toggle');
+    $('.js-show-facet').on('change', function(){
+      var section = $(this).val();
       window.cards.setFacet(section);
     });
   }
--- a/public/javascript/ep-card-filter.js
+++ b/public/javascript/ep-card-filter.js
@@ -230,6 +230,14 @@ $(document).ready(function(){
         $referenceCard.offset().top - referenceCardPixelsIntoViewport
       );
 
+      // Some mobile browsers (eg: iOS Safari) make it really difficult to
+      // unfocus a `select` element. This is a pain for us, because it means
+      // our "temporary" absolute positioning fix (.js-fixed-child--absolute)
+      // could remain in place much longer than required. Manually blurring
+      // the input once a selection has been made avoids this problem, and
+      // most users won't even noticed it's happened.
+      $(this).trigger('blur');
+
     });
   }
 
--- a/public/javascript/ep-card-filter.js
+++ b/public/javascript/ep-card-filter.js
@@ -200,8 +200,36 @@ $(document).ready(function(){
     });
 
     $('.js-show-facet').on('change', function(){
+      // Changing the facet has the potential to resize every card on the
+      // page. If cards above the current viewport are resized, you can easily
+      // lose your place in the list. So we're going to do some devious maths
+      // to pick a "reference card" (usually the top left visible card) and
+      // make sure that the viewport scrolls to maintain that reference card's
+      // position on screen, even if all the preceding cards change size.
+      var pixelsAboveViewport = $(window).scrollTop();
+      var referenceCardPixelsIntoViewport;
+      var $referenceCard;
+
+      // Find the reference card.
+      $('.person-card').each(function(){
+        var thisCardOffset = $(this).offset().top;
+        if ( thisCardOffset > pixelsAboveViewport ) {
+          $referenceCard = $(this);
+          referenceCardPixelsIntoViewport = thisCardOffset - pixelsAboveViewport;
+          return false; // break out of .each loop
+        }
+      });
+
+      // Actually set the visible facet on all the cards.
       var section = $(this).val();
       window.cards.setFacet(section);
+
+      // Now the facet has changed, adjust the window's scroll position so the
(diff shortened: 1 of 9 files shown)


views/sass/_components.scss
```diff
@@ -180,6 +180,21 @@ p.lead {
     }
 }
 
+.form-control,
+input[type].form-control {
```
zarino: I'm having to overload selectors with this `input[type]` stuff because the `input[type="…"]` selectors in `_typography.scss` are just way too strong.

Bootstrap gets round this by not applying any styles to inputs by default, instead forcing you to add a `.form-control` class if you want your inputs styled. We could do the same. But I wonder whether there’s a happy medium?

Any thoughts @wrightmartin @ondenman @octopusinvitro?

views/term_table.erb
```diff
@@ -107,130 +99,146 @@
...
+                                    <option value="contacts">Contact details</option>
+                                    <option value="identifiers">Identifiers</option>
+                                </select>
+                            </span>
+                        </p>
+                    </div>
                 </div>
-            </div>
 
-            <div class="grid-list" data-active-section="bio">
```
zarino: Sorry it appears I've changed lots of lines in this file – in actuality, all I did was indent it all 4 spaces further, so I could wrap it in the `<div class="js-fixed-parent">`.

zarino:
Would one (or more!) of @octopusinvitro @ondenman @chrismytton like to review this?

tmtmtmtm:
@zarino I'm assuming the big "What's in this dataset" section shouldn't still be saying "Click one of the categories to get a taste of the information we have so far"?

zarino:
> @zarino I'm assuming the big "What's in this dataset" section shouldn't still be saying "Click one of the categories to get a taste of the information we have so far"?

Good catch!

tmtmtmtm:
@zarino: generally looks great, yay! This is a *lot* better than what we currently have. I'm finding it a little bit jumpy at times though. e.g. if I'm on https://everypolitician-viewe-pr-14341.herokuapp.com/estonia/riigikogu/term-table/12.html, and scrolled to the Annely Akkermann row:

screen shot 2016-08-10 at 16 45 58

when I switch to the 'Social links' or 'Contact details' view , that row mostly shifts upwards out of sight again:

screen shot 2016-08-10 at 16 46 24

Identifiers then brings them down again:

screen shot 2016-08-10 at 16 47 53

zarino:
@tmtmtmtm: Yeah, when you switch to showing different bits of the cards, there's no guarantee the cards will be the same height (chances are, they definitely won't) so they'll move around, which can accumulate to quite a large jump when you're scrolled a long way down the list.

@tmtmtmtm: What did you expect to happen when you changed the "show details" option? Perhaps that the cards currently at the top of your view stayed at the top of your view?

tmtmtmtm:
@zarino Yeah — I would have expected the top of my screen to stay consistent, so if I start at 

screen shot 2016-08-10 at 17 05 44

and switch to Social, I'd be at:

screen shot 2016-08-10 at 17 05 55

If that's in "Eeek! Way too difficult" territory, I'd also be happy enough with it snapping to the top of those cards, like:

screen shot 2016-08-10 at 17 06 04

But the current behaviour pushes us down to the next set, which I find quite odd:

screen shot 2016-08-10 at 17 06 30

It's not a show-stopper, but it was quite jarring.

zarino:
That should be possible. I'll take a look tomorrow.

zarino:
@tmtmtmtm – It now attempts to maintain the visual scroll position after the "show details" select box is changed. Works in all the browsers I’ve tested it in, apart from:

* IE8 – doesn't support `position: fixed` so the "show details" select can't follow you down the page, meaning nothing to worry about here!
* IE9 – some weird flexbox CSS bug is causing the person-cards not to wrap onto multiple lines, meaning you can't scroll down the page at all – I'll try to fix that later.
* Safari on Mac – there's a few milliseconds between the CSS styles being updated and the javascript resetting the scroll position. It's noticeable that something has flashed on the screen. But I'm guessing this is better than losing your place in the list?
* Safari on iOS – Oh, fun and games!! We’re falling foul of [this "feature" in Safari iOS](http://stackoverflow.com/questions/29001977 ), which pulls you up to the top of the page when you focus the `position: fixed` select box. Another thing to fix tomorrow.

zarino:
I've fixed the iOS bug, so this is ready for review. @chrismytton could you review the JavaScript? And @wrightmartin the Sass, maybe?

public/javascript/ep-card-filter.js
```diff
@@ -194,9 +199,45 @@ $(document).ready(function(){
       window.cards.saveState();
     });
 
-    $('[data-section-toggle]').on('click', function(){
-      var section = $(this).attr('data-section-toggle');
+    $('.js-show-facet').on('change', function(){
```
chrismytton: This line can be simplified by using `.change(handler)`, which is a [shortcut for `.on('change', handler)`](https://api.jquery.com/change/).

public/javascript/main.js
```diff
@@ -272,13 +268,166 @@
...
+      } else {
+        // $parent is not in view. Unfix the child.
+        $spacer.hide();
+        $child.removeClass('js-fixed-child--fixed').css({
+          width: ''
+        });
+      }
+    }
+
+    updateChildPositionAbsolute = function updateChildPositionAbsolute($parent, $child, $spacer){
```
chrismytton: I think you're missing a `var` at the start of this line.

public/javascript/ep-card-filter.js
zarino: Is it not preferable to maintain consistency with all the other event bindings elsewhere, which all use `.on()`?

chrismytton:
@zarino Couple of comments on the js, but generally looking fine to me! 👍  🎉

public/javascript/ep-card-filter.js
chrismytton: I can only see two others in this file, but I guess that does make it the majority 😉  

I agree, consistency is best for this pull request, we can change to the shorthand methods in a future PR 👍

zarino: > we can change to the shorthand methods in a future PR

Interesting. My vote would always be for using `.on()` everywhere – although I'd be interested to hear the argument for shorthand methods, aside from being five characters shorter.

zarino:
@chrismytton @tmtmtmtm: I’ve rebased this on top of the latest master. Ran the tests locally in my vm, and they all passed, and it all looks and works fine when I poke at it in the browser. So I think we’re ready to merge this in, if you’re happy with it?

tmtmtmtm:
The travis failures here look like they're from https://github.com/everypolitician/everypolitician-data/pull/18231#issuecomment-252073921

I've fixed that in https://github.com/everypolitician/everypolitician-data/pull/18255, and will try restarting travis here when that gets merged.

zarino:
> Is this fairly easy to fix, or is it likely to be quite involved? I think my expected behaviour would be that a search would always bring the first matches for that search to the top

Shouldn't be hard to fix. I'll take a look now.

> making the search box much more prominent like this also makes #10276 even more important now

Good spot! I will fix that at the same time.

zarino:
@tmtmtmtm – Those two issues have been fixed now. Modified keys (eg: Ctrl-A) now work as you’d expect in the search box (fixing #10276) and searching takes you back to the top of the card list, so you don't get stranded at the bottom.

tmtmtmtm:
Woot! This is a huge improvement. Thanks @zarino and @octopusinvitro

PR #14401, opened 2016-08-12
tidy indenting of JavaScript in main.js

davewhiteland:
Just nudging javascript one notch inwards, within the jQuery "document ready" call.

Also made the minority places where there was 4-space indenting use 2-space to match the style in the rest of the file.

Tidying this up because I want to make changes in `main.js` (for #11911) and don't want to fix this up as part of that issue.

### DIFF
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -387,47 +387,47 @@ $(function(){
       }
   })
 
-// Google Events Tracking
+  // Google Events Tracking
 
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
https://github.com/everypolitician/viewer-sinatra/pull/14401/files?w=1 makes it much clearer that there's no substantial change here, so 👍

Issue #13599, opened 2016-07-25
One download page per legislature


zarino:
# Problem

We currently have a download page per country, but all of the download options are separated by legislature. This has a few drawbacks:

* Only the first legislature is visible on pageload, and as we add more historical data to EP, the legislatures after it will only get further down the page. On first page load, it's not obvious at all that the page contains data for more than one legislature.
* The need to pass a legislature name and a term id to the download page has resulted in somewhat overcomplicated code like https://github.com/everypolitician/viewer-sinatra/pull/12987

# Proposal

Create a download page per legislature. 

<del>Link between them via the new tab component introduced in https://github.com/everypolitician/viewer-sinatra/issues/13598.</del> Broken into separate issue: #15240 

# Acceptance Criteria

Clicking the "Download data" button on a legislature’s term_table page, should take me to a download page showing *only* the legislature I came from.

<del>In addition, if the legislature has lots of historical data, the data for the term I came from should be highlighted above any others.</del> Exists as separate issue: #14427

# Not Required

n/a

# Prerequisites

Having https://github.com/everypolitician/viewer-sinatra/pull/13598 merged would give us a tab component we could re-use for the navigation between legislature download pages. (Also useful as an above-the-fold way of signposting that we *even have* data for other legislatures in the same country).

# Related Issues

Came up as a solution to https://github.com/everypolitician/viewer-sinatra/pull/12987

# Due Date

n/a

# Design?

Will need some design time, to lay out the new single-legislature download pages.

# Text?

n/a

# Bloggable?

n/a

tmtmtmtm:
Done

PR #14425, opened 2016-08-12
add fragment id (#) to download button href, targeting specific legislature

### LINKED ISSUE #14423: jump to the correct legislature on the download page
# Problem

When clicking on the download data button (to go to the download page), jump to the _correct_ legislature on the download page if there is more than one.

This is the first part of #11911, broken out for clarity.

# Proposed Solution

Add an anchor tag for the legislature, and a matching target on the download page (house slug seems ideal).

# Acceptance Criteria

Clilcking on a _download data_ button goes to the appropriate legislature on the download page for bicameral systems; where there is only one, it should simply go to that one.

# Not Required

No need to worry about specific terms; that's a separate issue

# Prerequisites

None

# Related Issues

See #1191, and PR #12987.

# Due Date

N/A

# Design?

No

# Text?

No 

# Bloggable?

No

davewhiteland:
Adds anchor tags to download buttons to ensure that when the user lands on the download page, they do so at the correct legislature (which matters in bicameral systems, that is, if there is more than one).

Legislature-specific download buttons occur here:

* each country page (where available terms are _listed_)
* each term table page

Closes #14423 (which is a distillation of the first part of #11911).

Note the targets for these anchor tags (`id`) are already in place, so adding the # to the URL is sufficient for this to work.

### DIFF
--- a/views/country.erb
+++ b/views/country.erb
@@ -13,7 +13,7 @@
         <div class="country__legislature__header">
           <h3><%= house[:name] %></h3>
           <!-- download link as button? -->
-          <a class="button button--quarternary" href="download.html">
+          <a class="button button--quarternary" href="download.html#legislature-<%= house[:slug].downcase %>">
             <i class="fa fa-download"></i>
             Download <span class="large-screen-only">data</span>
           </a>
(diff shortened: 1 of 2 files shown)


davewhiteland:
Closing this because #13599 (one download page per legislature) is a better solution.

PR #14505, opened 2016-08-14
Start extracting logic to Page classes

tmtmtmtm:
Viewer-Sinatra started as a Quick-and-Dirty approach to throwing together a few simple pages, but we've stretched that long beyond breaking point. It’s time we tidied things up a bit.

https://github.com/everypolitician/viewer-sinatra/pull/5625 was a step towards that, but I’m not so sure that that’s going on the right direction, so I’ve tried rethinking the deeper approach we’re taking.

Due to the slightly idiosyncratic nature of how the site is built (and deployed, as a spidered static version), we don't need lots of the more complex behaviour that has driven the evolution of most modern web frameworks, so we should resist the lure of adding complexity, abstraction, and indirection until we actually need it, rather than because we think we _will_ need it.

It’s also important that we evolve out of what we currently have, rather than completely rearchitect around some other paradigm. We want to stay as lightweight as possible, but whilst introducing a better separation of concerns.

My proposal, for which this PR implements a few examples, whilst allowing all the other pages to be gradually migrated, is as follows:

---

The business logic of the site should live in a series of `Page` classes.  Page objects themselves should be entirely unaware of the outside world in either direction.

That means they should know nothing about the environment from which they've been called. Not only should they should have no awareness of HTTP, query variables, etc, but equally importantly they should know nothing about Sinatra either. If we were ever to switch to a different framework, these classes should not need to change.

Similarly, they should know nothing about the Template engine that they will be passed to. Switching between ERB and HAML, for example, should not require any changes the Page classes. (The one concession we make here is that we assume that any templating system we use will have the ability to call methods on objects.)

The routes in the Sinatra app should be as thin as possible, brokering the input and output by instantiating the relevant Page object, based on the request, and then usually handing that off to a templating engine to produce the output.

The only thing passed to a template should be the Page object, the public methods of which will provide all the data required. As such, every Page should have a test script that tests every public method on it.

---

Decoupling the various parts like this should allow us to write better business-logic level code (and test it much more easily). Further abstraction should arise naturally out of normal development practice of having clean, maintainable code in the Page classes, rather than out of an underlying philosophy of how webapps _should_ work.

Migrating each route is then largely a matter of extracting the code for setting each instance variable into a method on the relevant Page class instead, testing all those methods, and changing the template to call the methods rather than access the variables directly. We can do some light refactoring as we go, but anything complex should happen separately from the extraction.


chrismytton:
Awesome, this is a big improvement, easier to test and easier to understand than a big `app.rb` with a bag of helper methods. Big 👍  from me! 🚀

PR #14535, opened 2016-08-15
when download data is clicked on term table, highlight the term

### LINKED ISSUE #14427: highlight desired term on download page
# Problem

When clicking *download* on a term table page, the download page may show many terms. Since we know which term the user has "come from", we should highlight the desired one.

This was the first half of the second part of #11911. Obviously this is more granular, and simpler. 
# Proposed Solution

Use JS to intercept clicks on the download button, and inject query vars for the desired term and house. (The house is already _implied_ by the # fragment identifier in the URL, but by passing data as a query var, I don't think we should expect the target page to _also_ inspect the anchor tag to infer the house).

A working solution has been fundamentally implemented in PR #14425 but in such a stuttering, meandering way, it's going to be easiest to cherry-pick (or re-code) from there as a new PR, decoupled from the logic that also hides terms when there are many.

# Acceptance Criteria

Clicking on the *download data* button on a term table goes to a download page on which the correct term is highlighted (subtle yellow background (CSS class) is enough for now).

Arriving at the download page any other way (e.g., from the country-level page) highlights no term.

Note: if no JS, this functionality, which is an interface enhancement, is simply not present. The link still goes to the correct legislature (because that's handled by anchor tag: #14425 

# Not Required

N/A

# Prerequisites

None (not using the anchor tag, so this is independent of #14425 

# Related Issues

This was the first half of the second part of #11911. Simpler.

# Due Date

N/A

# Design?

Uses CSS class for highlight, but that can be changed if necessary by another issue for the designers. So for now, suggest simple CSS class with inoffensive (pale custard yellow/feeble highlighter pen) background colour.

# Text?

None

# Bloggable?

No

davewhiteland:
Closes #14427 

This code replaces the existing JavaScript (deployed but not being used because the download button is not currently passing term (and house) as query variables) with neater code.

It's solving the problem of clicking download from a term table page, and not being able to see that term amongst the terms listed on the download page in the case where there are too many: the solution is to reduce the number of terms displayed, and highlight the specific term.

The business of displaying only a few terms on the download page is now better handled by a `collapseDisplayedItems` function in `main.js`. The code is cleaned up with suggestions and review from @zarino and @tmtmtmtm in previous PR #12987, which this PR supersedes.

The `collapseDisplayedItems` function hides all but `maxDisplayItems` terms, if there are more than `minThresholdItems`. If there's a selected term (i.e., after a click on *Download data* from a term page) it is included in the displayed items.

Both `term` and `house` (as slugs) are passed from the term download button; if either is lacking (no term is selected) then multiple terms are still collapsed on the download page, but no specific term is hidden. By default, most recent terms are shown.

The highlighted term is given CSS class `download-terms__highlighted`.


public/javascript/main.js
```diff
@@ -479,3 +479,8 @@ $(function(){
 
   }
 });
+
+function getQueryParam(key) {
```
tmtmtmtm: is this something you wrote yourself, or is it taken from somewhere else? (and if the latter should it have attribution?)

public/javascript/main.js
```diff
@@ -385,7 +385,21 @@ $(function(){
...
+
+  $(".js-download-with-term").on("click", function(e){
+    var term = $(e.target).data("term");
+    var house = $(e.target).data("house");
+    if (term && house) {
+      // inject query vars into URL before the target anchor
+      e.preventDefault();
+      var href = $(e.target).attr('href');
+      if (href.indexOf('#') == -1) {
+        href = href + "#"; // add anchor tag at end of URL
```
tmtmtmtm: Comments that simply state in English what a fairly simple piece of code is doing are usually a red flag. Perhaps extract this to a suitably named helper function instead?

public/javascript/main.js
```diff
@@ -478,9 +492,61 @@ $(function(){
...
+// collapseDisplayedItems:
+// If there are too many items being displayed in the list, hide
+// all but a few, making sure the target element (if there is one)
+// is included in those that are not hidden. Add a button for
+// revealing the hidden items.
+// Note this is used on the download page, where (currently) there
+// may be more than one such list (e.g., one for each legislature).
+
+function collapseDisplayedItems(
+    $ul,               // unordered list elementthat needs to be collapsed
```
tmtmtmtm: space missing between 'element' and 'that'

public/javascript/main.js
```diff
@@ -478,9 +492,61 @@ $(function(){
...
+                       // note: <ul> must have an id
```
tmtmtmtm: It's probably not worth actually fixing it unless it's simple, but this seems like a slightly odd restriction. My JS isn't really good enough, but is there a way to do this with a closure around `$ul` instead?

public/javascript/main.js
```diff
@@ -478,9 +492,61 @@ $(function(){
...
+      upperBound = maxUpperBound;
+      lowerBound = lowerBound - delta;
+    }
+    if (lowerBound < 0) {
+      upperBound = Math.min(upperBound + Math.abs(lowerBound), maxUpperBound);
+      lowerBound = 0;
+    }
+    $listItems.slice(lowerBound, upperBound).removeClass(hiddenClassName);
+
+    $('.' + hiddenClassName).hide();
```
tmtmtmtm: Does this means that the `hiddenClassName` needs to be something that is unique to these elements? Should this be restricted to only being against the elements under `$ul`?

public/javascript/main.js
```diff
@@ -478,9 +492,61 @@ $(function(){
...
+    minThresholdItems, // only collapse if more elements than this
+    maxDisplayItems,   // number of elements to show when collapsed
+    buttonText         // message on reveal button
+  ) {
+  var $listItems = $ul.find('li');
+  if ($listItems.length > minThresholdItems) {
+    $listItems.addClass(hiddenClassName);
+    // want a slice around targetIndex of maxDisplayItems
+    var targetIndex = $listItems.index($targetItem);
+    var maxUpperBound = $listItems.length;
```
tmtmtmtm: I'm finding this logic a little too difficult to follow, and it interrupts the flow of this function quite severely. This seems like it should really be a separate function that either partitions a list for you, or even perhaps just one that gives you the indices you want. (It also seems much more complex that I would expect. Is that simply "because Javascript"?)

tmtmtmtm:
Some comments in-place. I'm a little wary of diminishing returns here, so 'TODO' comments in-place are also plausible options for some of these things…

public/javascript/main.js
davewhiteland: yes, using the .next() sibling selector instead is a better solution: since the toggle button is being
injected directly above the list it relates to, so can assume that the list is its next sibling (instead of using ID).

davewhiteland:
invitation for review @tmtmtmtm

davewhiteland:
It's worth mentioning perhaps that now it's neatly encapsulated, it becomes clear that the implementation I ought to want is `outerList` not `innerList`... because really this is just to add the CSS class to the items that's _are *not* in the inner list_ (because those are the ones that need to respond to show/hide toggling).

tmtmtmtm:
Reverting back to @davewhiteland to work out what should happen now that this page has been split into per-legislature rather than per-country.

PR #14572, opened 2016-08-15
Extract needed page

ondenman:
Extracted logic in /needed to new class. 
Added VCR and made it work only for requests made in tests for this class. (Requests made to DATASOURCE url ''https://cdn.rawgit.com/everypolitician/everypolitician-data...' are not wrapped in VCR.use_cassette blocks.)
Referenced new class in app.rb and erb.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -200,7 +200,7 @@ get '/status/all_countries.html' do
 end
 
 get '/needed.html' do
-  @page = Page::Needed.new
+  @page = Page::Needed.new(ENV['GITHUB_ACCESS_TOKEN'])
   erb :needed
 end
 
--- a/app.rb
+++ b/app.rb
@@ -200,18 +200,7 @@ get '/status/all_countries.html' do
 end
 
 get '/needed.html' do
-  if (token = ENV['GITHUB_ACCESS_TOKEN']).to_s.empty?
-    logger.warn 'No GITHUB_ACCESS_TOKEN found'
-    client = Octokit::Client.new
-  else
-    client = Octokit::Client.new(access_token: token)
-  end
-  client.auto_paginate = true
-
-  @to_find   = client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Find'
-  @to_scrape = client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Scrape'
-  @to_finish = client.issues 'everypolitician/everypolitician-data', labels: 'New Country,3 - WIP'
-
+  @page = Page::Needed.new
   erb :needed
 end
 
--- a/app.rb
+++ b/app.rb
@@ -200,7 +200,7 @@ get '/status/all_countries.html' do
 end
 
 get '/needed.html' do
-  @page = Page::Needed.new(ENV['GITHUB_ACCESS_TOKEN'])
+  @page = Page::Needed.new(access_token: ENV['GITHUB_ACCESS_TOKEN'])
   erb :needed
 end
 
--- a/lib/page/needed.rb
+++ b/lib/page/needed.rb
@@ -1,8 +1,10 @@
+require 'octokit'
+
 module Page
   class Needed
-    def initialize(access_token = '')
+    def initialize(access_token: required)
+      token_error if access_token.to_s.strip.empty?
       @access_token = access_token
-      warn 'No GITHUB_ACCESS_TOKEN found' if @access_token.to_s.empty?
     end
 
     def to_find
@@ -18,6 +20,10 @@ module Page
     end
 
     private
+
+    def required
+      raise ArgumentError, 'No GITHUB_ACCESS_TOKEN provided'
+    end
 
     def client
       client = Octokit::Client.new(access_token: @access_token)
--- a/lib/page/needed.rb
+++ b/lib/page/needed.rb
@@ -7,20 +7,24 @@ module Page
     end
 
     def to_find
-      client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Find'
+      issues 'New Country,To Find'
     end
 
     def to_scrape
-      client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Scrape'
+      issues 'New Country,To Scrape'
     end
 
     def to_finish
-      client.issues 'everypolitician/everypolitician-data', labels: 'New Country,3 - WIP'
+      issues 'New Country,3 - WIP'
     end
 
     private
 
     attr_reader :access_token
+
+    def issues(labels)
+      client.issues 'everypolitician/everypolitician-data', labels: labels
+    end
 
     def client
       client = Octokit::Client.new(access_token: @access_token)
--- a/lib/page/needed.rb
+++ b/lib/page/needed.rb
@@ -1,0 +1,28 @@
+module Page
+  class Needed
+    def to_find
+      client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Find'
+    end
+
+    def to_scrape
+      client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Scrape'
+    end
+
+    def to_finish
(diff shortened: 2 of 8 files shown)


lib/page/needed.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def to_finish
+      @client.issues 'everypolitician/everypolitician-data', labels: 'New Country,3 - WIP'
+    end
+
+    private
+
+    def make_client
+      if (token = ENV['GITHUB_ACCESS_TOKEN']).to_s.empty?
```
tmtmtmtm: The Page classes shouldn't be aware of the outside world, which includes ENV. You should pass this to the constructor instead.

lib/page/needed.rb
```diff
@@ -0,0 +1,31 @@
...
+      @client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Scrape'
+    end
+
+    def to_finish
+      @client.issues 'everypolitician/everypolitician-data', labels: 'New Country,3 - WIP'
+    end
+
+    private
+
+    def make_client
```
tmtmtmtm: Rather that making this a command that you call from the constructor to set an instance variable, it would be cleaner to just have it be a simple `client` method that creates it on demand and returns it, and make all the calls go through the method, rather than access the variable.

t/page/needed.rb
```diff
@@ -0,0 +1,39 @@
...
+require 'octokit'
+require 'pry'
+require 'vcr'
+
+describe 'Needed' do
+  VCR.configure do |config|
+    config.cassette_library_dir = 't/fixtures/vcr_cassettes'
+    config.hook_into :webmock
+    config.ignore_request do |req|
+      req[:uri].include? 'https://cdn.rawgit.com/everypolitician/everypolitician-data'
```
tmtmtmtm: Why is this required? As far as I can see none of the requests are hitting anything that matches this.

lib/page/needed.rb
```diff
@@ -0,0 +1,31 @@
+module Page
+  class Needed
+    def initialize
+      make_client
```
tmtmtmtm: `make_client` has two branches in it, but every client you're returning here seems to only follow one of those paths.

t/page/needed.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  it 'should return a list of countries labeled to_find' do
+    VCR.use_cassette('octokit - test to_find') do
+      list = subject.to_find
+      labels = ['New Country', 'To Find']
+      labels.must_include list.first[:labels].first[:name]
+    end
+  end
+
+  it 'should return a list of countries to_scrape' do
```
tmtmtmtm: Perhaps cleared to name this more explicitly as *not* returning any countries?

t/page/needed.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  subject do
+    Page::Needed.new
+  end
+
+  it 'should return a list of countries labeled to_find' do
+    VCR.use_cassette('octokit - test to_find') do
+      list = subject.to_find
+      labels = ['New Country', 'To Find']
+      labels.must_include list.first[:labels].first[:name]
```
tmtmtmtm: I'm finding this test a bit too confusing to follow. As this is running against a canned response, perhaps it would be cleared to test something like `list.map(&:title).include? 'Eritrea'` ?

t/page/needed.rb
```diff
@@ -0,0 +1,39 @@
...
+    config.ignore_request do |req|
+      req[:uri].include? 'https://cdn.rawgit.com/everypolitician/everypolitician-data'
+    end
+  end
+
+  subject do
+    Page::Needed.new
+  end
+
+  it 'should return a list of countries labeled to_find' do
```
tmtmtmtm: This (both in name, and in what it's actually testing) is a little bit too coupled to the internal implementation of the method. What you should be testing is that we can get a list of the countries that we still don't even need a source for. That is currently achieved by querying labels, but that should be irrelevant to the test. If we entirely change how that works (e.g. by looking them up in a Google Spreadsheet) we should only need to change the code — not the tests.

app.rb
```diff
@@ -203,18 +203,7 @@ def soft_redirect(url, page_title)
...
-  else
-    client = Octokit::Client.new(access_token: token)
-  end
-  client.auto_paginate = true
-
-  @to_find   = client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Find'
-  @to_scrape = client.issues 'everypolitician/everypolitician-data', labels: 'New Country,To Scrape'
-  @to_finish = client.issues 'everypolitician/everypolitician-data', labels: 'New Country,3 - WIP'
-
+  @page = Page::Needed.new(ENV['GITHUB_ACCESS_TOKEN'])
```
tmtmtmtm: I have a very strong suspicion that we should always be passing named parameters (or Hash), rather than a list of contextless arguments. We don't necessarily need to do that in this PR; more just capturing the thought here…

lib/page/needed.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    def to_finish
+      client.issues 'everypolitician/everypolitician-data', labels: 'New Country,3 - WIP'
+    end
+
+    private
+
+    def client
+      client = Octokit::Client.new(access_token: @access_token)
```
tmtmtmtm: This is a slight change in behaviour from the original. Previously if `access_token` was empty, the call would be a plain `Octokit::Client.new`, whereas you're now calling `Octokit::Client.new(access_token: '')`

That _seems_ to Do The Right Thing, but it's not obvious from the Octokit docs that it's guaranteed to. 

However, running this from here _without_ an access token currently fails with "403 - API rate limit exceeded" — which we're not capturing.

As in production we always provide an access token anyway, I suspect it might make things much simpler if we just declare that it's compulsory rather than optional…

lib/page/needed.rb
```diff
@@ -0,0 +1,38 @@
...
+
+module Page
+  class Needed
+    def initialize(access_token: required)
+      token_error if access_token.to_s.strip.empty?
+      @access_token = access_token
+    end
+
+    def required
+      token_error
```
tmtmtmtm: This is an interesting approach, but I'm not sure what you gain from the indirection of a `required`  method _and_ a `token_error` method, rather than just one or the other.

lib/page/needed.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def token_error
```
tmtmtmtm: Neither this nor `required` seem like they should be `public`.

t/page/needed.rb
```diff
@@ -0,0 +1,37 @@
...
+
+describe 'Needed' do
+  VCR.configure do |config|
+    config.allow_http_connections_when_no_cassette = true
+    config.cassette_library_dir = 't/fixtures/vcr_cassettes'
+    config.hook_into :webmock
+  end
+
+  subject do
+    Page::Needed.new(access_token: ENV['GITHUB_ACCESS_TOKEN'])
```
tmtmtmtm: You're still examining the environment in the test, which means these tests fail when I run them, as I don't have a GITHUB_ACCESS_TOKEN set. 

You also don't seem to have a test for the case where someone doesn't have it set, making sure that it throws the correct Error.

t/page/needed.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  it 'should return a list of countries to find' do
+    VCR.use_cassette('octokit - test to_find') do
+      list = subject.to_find
+      list.map(&:title).include? 'Eritrea'
+      list.map(&:title).include? 'Cuba'
+      list.map(&:title).include? 'Qatar'
```
tmtmtmtm: What is the scenario where the two prior tests could pass but this one fail?

ondenman:
@tmtmtmtm Added fixup to remove VCR cassettes. Will need to squash before merge.

PR #14612, opened 2016-08-16
Split the exhaustive web tests to a separate rake task

tmtmtmtm:
Running the entire test suite takes forever because it has to make sure
that every country works as a fully rendered page.

If we split those out to a separate `rake test:web` target, and ensure that
`rake test:all` runs all of them, then we can still run the full test suite
when needed, but the day-to-day run of the Page tests can be lightning
fast.

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -2,7 +2,21 @@ require 'rake/testtask'
 require 'rubocop/rake_task'
 
 Rake::TestTask.new do |t|
-  t.pattern = 't/**/*.rb'
+  t.warning = true
+  t.test_files = FileList['t/page/*.rb']
+end
+
+Rake::TestTask.new do |t|
+  t.name = 'test:web'
+  t.warning = false
+  t.verbose = true
+  t.test_files = FileList['t/web/*.rb']
+end
+
+Rake::TestTask.new do |t|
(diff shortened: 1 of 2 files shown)


Rakefile.rb
```diff
@@ -2,7 +2,21 @@
 require 'rubocop/rake_task'
 
 Rake::TestTask.new do |t|
-  t.pattern = 't/**/*.rb'
+  t.warning = true
+  t.test_files = FileList['t/page/*.rb']
+end
```
chrismytton: It might be worth adding the (undocumented, as far as I can tell) `t.description = 'Blah'` option here.

That would make the output of `rake -T` a bit more useful:

```
$ be rake -T
rake rubocop               # Run RuboCop
rake rubocop:auto_correct  # Auto-correct RuboCop offenses
rake test                  # Run tests
rake test:all              # Run tests
rake test:web              # Run tests
```

chrismytton:
One small comment about adding the `t.description` option to the test tasks, but otherwise this looks good to me! 👍

Rakefile.rb
tmtmtmtm: oooh. I was unaware of that!

PR #14613, opened 2016-08-16
Extract country page class

octopusinvitro:
This pull request extracts part of the logic under the `/:country/` route into its own class `Country`.

* The relevant templates were updated to use the member fields of the class instance.

* The main layout template had to be updated since it was using `@page_title`, which we don't need to pass to the template as the `Country` class has its own `title` member. So now, if we pass a `@page` instance and the `title` is available, we use it. The `@page_title` is used to render the title of the page in the browser.

* Since we also don't have a `@country` when we are passing a `@page`, the same had to be done so that the main template layout uses the `country` field of the `@page` if it is available. The `@country` is used in the main layout template to render the heading of the page in the page's body.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -59,10 +59,10 @@ get '/countries.html' do
 end
 
 get '/:country/' do |country|
-  if @country = ALL_COUNTRIES.find { |c| c[:url] == country }
-    @page_title = "EveryPolitician: #{@country[:name]}"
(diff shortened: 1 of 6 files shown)


tmtmtmtm:
replaced by https://github.com/everypolitician/viewer-sinatra/pull/14617

PR #14617, opened 2016-08-16
Extract a Country page class

octopusinvitro:
This pull request extracts the logic under the first `/:country/` route into a class `Country`.

* The relevant templates were updated to use the member fields of the class instance.

* The main layout template had to be updated since it was using `@page_title`, which we don't need to pass to the template as the `Country` class has its own `title` member. So now, if we pass a `@page` instance and the `title` is available, we use it. The `@page_title` is used to render the title of the page in the browser.

* Since we also don't have a `@country` when we are passing a `@page`, the same had to be done so that the main template layout uses the `country` field of the `@page` if it is available. The `@country` is used in the main layout template to render the heading of the page in the page's body.

----

Test page: https://everypolitician-viewe-pr-14617.herokuapp.com/british-virgin-islands/

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -58,7 +58,7 @@ get '/countries.html' do
 end
 
 get '/:country/' do |country|
-  @page = Page::Country.new(country, ALL_COUNTRIES)
+  @page = Page::Country.new(country)
   pass unless @page.country
   erb :country
 end
--- a/app.rb
+++ b/app.rb
@@ -58,7 +58,8 @@ get '/countries.html' do
 end
 
 get '/:country/' do |country|
-  @page = Page::Country.new(country)
+  SHA   = cjson.split('/')[-2]
+  @page = Page::Country.new(country, SHA)
   pass unless @page.country
   erb :country
 end
--- a/app.rb
+++ b/app.rb
@@ -58,8 +58,8 @@ get '/countries.html' do
 end
 
 get '/:country/' do |country|
-  @country = ALL_COUNTRIES.find { |c| c[:url] == country } || pass
-  @page_title = "EveryPolitician: #{@country[:name]}"
+  @page = Page::Country.new(country, ALL_COUNTRIES)
+  pass unless @page.country
   erb :country
 end
 
--- a/app.rb
+++ b/app.rb
@@ -58,8 +58,10 @@ get '/countries.html' do
 end
 
 get '/:country/' do |country|
-  SHA   = cjson.split('/')[-2]
-  @page = Page::Country.new(country, SHA)
+  @page = Page::Country.new(
+    country: country,
+    index:   EveryPolitician::Index.new(index_url: cjson)
+  )
   pass unless @page.country
   erb :country
 end
--- a/lib/page/country.rb
+++ b/lib/page/country.rb
@@ -1,22 +1,23 @@
+# frozen_string_literal: true
 require 'everypolitician'
 
 module Page
   class Country
-    def initialize(slug, ref = 'master')
-      @slug = slug
-      @ref  = ref
+    def initialize(country:, index:)
+      @slug  = country
+      @index = index
     end
 
     def country
-      @country ||= EveryPolitician::Index.new(ref).country(slug)
+      index.country(slug)
     end
 
     def title
-      "EveryPolitician: #{country.name}" if country
+      "EveryPolitician: #{country.name}"
     end
 
     private
 
-    attr_reader :slug, :ref
+    attr_reader :slug, :index
   end
 end
--- a/lib/page/country.rb
+++ b/lib/page/country.rb
@@ -2,10 +2,14 @@ require 'everypolitician'
 
 module Page
   class Country
-    attr_reader :country, :title
-    def initialize(country)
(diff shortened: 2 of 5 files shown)


lib/page/country.rb
```diff
@@ -0,0 +1,9 @@
+module Page
+  class Country
+    attr_reader :country, :title
+    def initialize(country, all_countries)
+      @country = all_countries.find { |c| c[:url] == country }
+      @title = "EveryPolitician: #{@country[:name]}" if @country
```
tmtmtmtm: Why are you cramming all this logic into `initialize`? These seem like they should just be normal methods.

app.rb
```diff
@@ -59,13 +59,14 @@ def soft_redirect(url, page_title)
 end
 
 get '/:country/' do |country|
-  @country = ALL_COUNTRIES.find { |c| c[:url] == country } || pass
-  @page_title = "EveryPolitician: #{@country[:name]}"
+  @page = Page::Country.new(country, ALL_COUNTRIES)
```
tmtmtmtm: I don't think you want to be passing things like ALL_COUNTRIES around. You should be able to just call `EveryPolitician.country(country)` in the Page class instead.

t/page/country.rb
```diff
@@ -0,0 +1,29 @@
...
+      {
+        name: 'Abkhazia',
+        url:  'abkhazia',
+      },
+    ]
+  end
+
+  it 'detects that a country is not missing' do
+    page = Page::Country.new('abkhazia', all_countries)
+    refute_nil page.country
```
tmtmtmtm: Presumably you'd quite like it to get the correct country here, rather than just anything that's non nil?

lib/page/missing_country.rb
```diff
@@ -0,0 +1,8 @@
+module Page
+  class MissingCountry
+    attr_reader :missing
+    def initialize(country, world)
+      @missing = world[country.to_sym]
```
tmtmtmtm: It would be more in keeping with the other classes if this was `@country`, rather than `@missing`. And, after all, the thing that this variable contains is actually country, not a "missing".

t/page/country.rb
```diff
@@ -0,0 +1,29 @@
...
+  let(:all_countries) do
+    [
+      {
+        name: 'Abkhazia',
+        url:  'abkhazia',
+      },
+    ]
+  end
+
+  it 'detects that a country is not missing' do
```
tmtmtmtm: This seems like a slightly odd primary purpose for the test. There's already a test below for handling the case where the country doesn't exist — this one seems like it should be about the normal case where you do get a country, and can then access something on that that's needed in the template (e.g. its legislatures)

tmtmtmtm:
This definitely seems cleaner than https://github.com/everypolitician/viewer-sinatra/pull/14613, but I think it's still passing too much around (the countries lists). Those aren't really things that vary per request, and are things we'd like to stop the routes having to know about. The main countries list is already encapsulated with `everypolitician-ruby`, so we should ideally just be using that directly (like in `Page::Download`)

The WORLD should probably evolve to have its own encapsulating object, but I suspect that handling that should really be in a separate PR anyway, and keep this one to just handing the first '/:country/' route, rather than both of them.

lib/page/country.rb
tmtmtmtm: `initialize` is doing a bit too much work here. I suspect I led you into this bad behaviour, though, with how `Page::Download` works, so I've cleaned that one up in https://github.com/everypolitician/viewer-sinatra/pull/14625

octopusinvitro: YUSSSSSS! :sparkles: :+1: :boom: 

Happinessssss  :notes:

lib/page/missing_country.rb
octopusinvitro: We thought about that as well, but the main layout template uses `@country` to display the headings (the name of the country, when you hit the country route), accessing `:name` and `:url`, but the missing country page doesn't display the country name (just the headings. We had to change that layout so that it could use the `@page.country` if it is available, i.e., for instances of a country page, but that would mean it would also use it when we send an instance of a missing country page... which is the reason why I think the old code contained `missing` instead of country.

We tried:
```erb
<% @country    = @page.country if (@page and @page.respond_to? :country and !@page.country.has? 'displayName' ) %>
```

in the old PR but decided to remove it because we were already feeling dirty enough with all those conditionals to add more on top of them, maybe we can revisit that line in this PR.

I think probably there is maybe too much logic in the layout template were the handling of the headings takes place, we could raise an issue to clean it.

tmtmtmtm: Ah, good point. But rather than working around that, I suspect this is another perfect place for making the change easy so you can make the easy change. I suggest we harmonise the interface of both `country` at least to the extent of the layout template being able to call what _it_ needs on either. 

As a step towards that I've broken out a World class in https://github.com/everypolitician/viewer-sinatra/pull/14627 to encapsulate working with that file (as we're going to need that anyway). I suspect we should next wrap up each key/value pair in that as a simple Struct with the required interface.

PR #14647, opened 2016-08-17
Let World return a suitable Country

tmtmtmtm:
We want to be able to plug a Country into the wrapper template so that
we can call `.name` and `.url` on it, without caring about whether it
originated from the `countries.json` or `world.json` files. So use a
Struct as a simple Adapter within the World class to allow us to call
these methods on the data from there.

### DIFF
--- a/lib/world.rb
+++ b/lib/world.rb
@@ -9,6 +9,17 @@ class World
     @_wjson ||= JSON.parse(File.read(file), symbolize_names: true)
   end
 
+  # super-simplistic adapter for the inner data. Over time it might make
+  # sense to properly extract this to its own class, but for now we just
+  # want to tidy the interface up a little so that this is substitutable
+  # for an `everypolitician-ruby` Country in a few well-defined places.
+
+  Country = Struct.new(:url, :name, :names)
+  def country(slug)
+    return unless found = as_json[slug.to_sym]
+    Country.new(slug, found[:displayName], found[:allNames])
+  end
+
   private
 
   attr_reader :file
--- a/t/helpers/world.rb
+++ b/t/helpers/world.rb
@@ -7,4 +7,13 @@ describe 'World' do
   it 'should know other names for countries' do
     subject.as_json[:estonia][:allNames].must_include 'Estland'
   end
+
+  it 'should give us countries as objects' do
+    subject.country('american-samoa').name.must_equal 'American Samoa'
+    subject.country('american-samoa').url.must_equal 'american-samoa'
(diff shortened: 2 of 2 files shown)


lib/world.rb
```diff
@@ -9,6 +9,17 @@ def as_json
     @_wjson ||= JSON.parse(File.read(file), symbolize_names: true)
   end
 
+  # super-simplistic adapter for the inner data. Over time it might make
+  # sense to properly extract this to its own class, but for now we just
+  # want to tidy the interface up a little so that this is substitutable
+  # for an `everypolitician-ruby` Country in a few well-defined places.
+
+  Country = Struct.new(:url, :name, :names)
```
chrismytton: Why is this `:url` and not `:slug`?

tmtmtmtm:
We have a little bit of a chicken-and-egg situation here in that the 'url' method should probably be 'slug', but we need to match the interface of the (adapted) countries.json to which we add a `url` as well. We _could_ clean [that](https://github.com/everypolitician/viewer-sinatra/blob/c748ec2bcfd011f998b7b85ff38ee56051c7b46b/app.rb#L24) up first, but then we'd need to change the templates too, and then we're out in the fields playing a gramophone at the bull.

This version seems like it will let us make https://github.com/everypolitician/viewer-sinatra/pull/14617 simpler, and then we can switch both versions to be `.slug` in one go at the end when everything is neatly harmonised.

lib/world.rb
tmtmtmtm: We have a little bit of a chicken-and-egg situation here in that the 'url' method should probably be 'slug', but we need to match the interface of the (adapted) countries.json to which we add a `url` as well. We _could_ clean [that](https://github.com/everypolitician/viewer-sinatra/blob/c748ec2bcfd011f998b7b85ff38ee56051c7b46b/app.rb#L24) up first, but then we'd need to change the templates too, and then we're out in the fields playing a gramophone at the bull.

This version seems like it will let us make https://github.com/everypolitician/viewer-sinatra/pull/14617 simpler, and then we can switch both versions to be `.slug` in one go at the end when everything is neatly harmonised.

chrismytton:
@tmtmtmtm Ah, I wan't aware of the [code setting `:url` to `:slug`](https://github.com/everypolitician/viewer-sinatra/blob/c748ec2bcfd011f998b7b85ff38ee56051c7b46b/app.rb#L24). In that case I think this is fine as an incremental step. So 👍  from me!

PR #14649, opened 2016-08-17
require_relative World

tmtmtmtm:
The app doesn't like us just 'require'-ing this directly. This is an
easy fix, but a little bit unsatisfying, not least because the tests
were passing without this, so it implies a deeper underlying problem
that we really need to address more generally.


tmtmtmtm:
a long discussion on the benefits of path-munging vs require-relative transpired on Slack. For now I'm just going to merge this so that things work again.

PR #14658, opened 2016-08-17
Extract house wikidata class

ondenman:
Extracted logic from `/:country/:house/wikidata` to new class.

Referenced new class in app.rb and era

### DIFF
--- a/lib/page/house_wikidata.rb
+++ b/lib/page/house_wikidata.rb
@@ -15,10 +15,6 @@ module Page
       country[:legislatures].find { |h| h[:slug].downcase == house_slug.downcase }
     end
 
-    def last_sha
-      house[:sha]
-    end
-
     def popolo
       popolo_file = EveryPolitician::GithubFile.new(house[:popolo], last_sha)
       JSON.parse(popolo_file.raw, symbolize_names: true)
@@ -31,5 +27,8 @@ module Page
     private
 
     attr_reader :house_slug, :country_slug
+    def last_sha
+      house[:sha]
+    end
   end
 end
--- a/lib/page/house_wikidata.rb
+++ b/lib/page/house_wikidata.rb
@@ -1,0 +1,31 @@
+require_relative '../../lib/popolo_helper.rb'
+
+module Page
+  class HouseWikidata
+    def initialize(slug, house)
+      @house = house
+      @slug = slug
+    end
+
+    def country
+      EveryPolitician.country(@slug)
+    end
+
+    def house
+      country[:legislatures].find { |h| h[:slug].downcase == @house.downcase }
+    end
+
+    def last_sha
+      house[:sha]
+    end
+
+    def popolo
+      popolo_file = EveryPolitician::GithubFile.new(house[:popolo], last_sha)
+      JSON.parse(popolo_file.raw, symbolize_names: true)
+    end
+
+    def page_title
+      "EveryPolitician: #{country[:name]} — #{house[:name]}"
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/page/house_wikidata.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def country
+      EveryPolitician.country(@slug)
+    end
+
+    def house
+      country[:legislatures].find { |h| h[:slug].downcase == @house.downcase }
+    end
+
+    def last_sha
```
tmtmtmtm: This appears to be a public method with no test.

t/page/house_wikidata.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  it 'should return a page_title' do
+    subject.page_title.must_equal 'EveryPolitician: Austria — Nationalrat'
+  end
+
+  it 'should return the country popolo' do
+    popolo = subject.popolo
+    keys = %i(posts persons organizations meta memberships events areas)
+    (popolo.keys & keys).must_equal keys
```
tmtmtmtm: I find this quite confusing. What are you actually testing here? Isn't this just a well disguised version of `popolo.keys.must_equal keys`? 

Aside from being expressed a little oddly, this isn't really testing anything particularly useful — just that the popolo looks like popolo. It might be better to test that it looks like the _right_ popolo — i.e at the sha you want — by testing something that will be fairly unique at that point.

lib/page/house_wikidata.rb
```diff
@@ -0,0 +1,31 @@
+require_relative '../../lib/popolo_helper.rb'
+
+module Page
+  class HouseWikidata
+    def initialize(slug, house)
+      @house = house
+      @slug = slug
+    end
```
tmtmtmtm: This reads quite oddly as the two instance variables look like they're at two different levels of specificity, although they're actually both the same. `slug` here is the *country* slug, and `house` is the *house* slug.

tmtmtmtm: Our style here is also to put a private `attr_reader` on the instance variables, and have all access go through the methods, rather than direct variable access.

lib/page/house_wikidata.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def country
+      EveryPolitician.country(@slug)
+    end
+
+    def house
+      country[:legislatures].find { |h| h[:slug].downcase == @house.downcase }
```
tmtmtmtm: Now that you've made country be an `EveryPolitician::Country` that could just be `country.legislature(house)`

lib/page/house_wikidata.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def last_sha
+      house[:sha]
+    end
+
+    def popolo
+      popolo_file = EveryPolitician::GithubFile.new(house[:popolo], last_sha)
+      JSON.parse(popolo_file.raw, symbolize_names: true)
+    end
+
```
tmtmtmtm: If you make `house` be an EveryPolitician::House, this could just be `house.popolo`

t/page/house_wikidata.rb
```diff
@@ -0,0 +1,22 @@
+require 'minitest/autorun'
+require_relative '../../lib/page/house_wikidata'
+
+describe 'HouseWikidata' do
+  subject do
+    Page::HouseWikidata.new('austria', 'Nationalrat')
```
tmtmtmtm: It's generally best to make the primary test be the way that something is actually used — i.e. 'nationalrat', rather than 'Nationalrat'. (If you need to test that it's case-insensitive, then you can add a second test that explicitly says that's what it's doing, but I don't think you need that here)

tmtmtmtm:
@ondenman I've updated this to the new style, and rebased over a not-yet-merged-because-travis branch. If this all looks good to you, I'll rebase properly again later.

PR #14659, opened 2016-08-17
Extract missing country page

octopusinvitro:
This PR extracts a missing country-page to be used by the :country route when information for a country is missing in everypolitician data.

The route was updated as well as the template that receives the `@page` instance from said route.

### DIFF
--- a/t/page/missing_country.rb
+++ b/t/page/missing_country.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'minitest/autorun'
 require_relative '../../lib/page/missing_country'
 require 'pry'
--- a/t/page/missing_country.rb
+++ b/t/page/missing_country.rb
@@ -4,12 +4,12 @@ require 'pry'
 
 describe 'MissingCountry' do
   it 'has a display name' do
-    page = Page::MissingCountry.new('eritrea')
+    page = Page::MissingCountry.new(country: 'eritrea')
     page.country.name.must_equal 'Eritrea'
   end
 
   it 'returns nil if country does not exist' do
-    page = Page::MissingCountry.new('narnia')
+    page = Page::MissingCountry.new(country: 'narnia')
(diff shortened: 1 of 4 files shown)


t/page/missing_country.rb
```diff
@@ -0,0 +1,15 @@
+require 'minitest/autorun'
+require_relative '../../lib/page/missing_country'
+require 'pry'
+
+describe 'MissingCountry' do
+  it 'a missing country has a display name' do
+    page = Page::MissingCountry.new('eritrea')
+    assert_equal page.country[:displayName], 'Eritrea'
```
tmtmtmtm: Again, I'd prefer we keep our syntax consistent across all the test files, rather than switching between `assert_equal` and `.must_equal`. (Open to persuasion that it should be `assert` style by default, but until then better to stay the same everywhere)

t/page/missing_country.rb
```diff
@@ -0,0 +1,15 @@
+require 'minitest/autorun'
+require_relative '../../lib/page/missing_country'
+require 'pry'
+
+describe 'MissingCountry' do
+  it 'a missing country has a display name' do
```
tmtmtmtm: a fairly tiny nit, but if we're going to use 'it' style, then it's good if it actually makes sense as a sentence. The context is already a Missing Country, so I think it's OK to just say "it 'has a display name'"

tmtmtmtm:
This looks good, but we've switched the preferred style out from under you, such that all Pages now need to be instantiated with an Index (and should use named parameters). That should finally fix all the test problems we had! This branch is currently passing on travis, so it _might_ be possible to merge now and then migrate to the new style separately, but I'm a little worried about doing that, so if you don't mind, I'd rather rebase this over master and do that before pushing this production. I'll do that later this morning once travis and rubocop start playing together more nicely (and you can keep going with the hydra instead!)

PR #14661, opened 2016-08-17
Remove test warnings

octopusinvitro:
We were having a warning when running tests, telling us that the `CJSON` and `SHA` variables were already set. This was because some tests were using them to set the data source for the tests (`Everypolitician.countries_json = CJSON % SHA`) and redefining them every time.

Even worse, this was making our suite be either randomly red or green depending if the tests were run in a certain order. EXAMPLE: #14678

I moved the declaration of the constants and the setting of `countries_json` to a test helprer, so that all tests can use it with the same values.

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -4,7 +4,7 @@ require 'rubocop/rake_task'
 Rake::TestTask.new do |t|
   t.warning = true
   t.description = 'Run "Page" tests'
-  t.test_files = FileList['t/page/*.rb', 't/helpers/*.rb']
+  t.test_files = FileList['t/test_helper.rb', 't/page/*.rb', 't/helpers/*.rb']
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
This doesn't actually seem to work at all now I'm afraid  — when I run this the data is just being loaded from `master` than the sha provided. (The fact that you had to "fix" this by just changing the expected values in the test, even though the data in https://github.com/everypolitician/everypolitician-data/blob/d8a4682f/countries.json can't have changed in the meantime, should have been a warning sign that something wasn't working right here…) 

I think the problem is at a higher level though — `everypolitician-ruby` shouldn't need to be configured via global state. (ping @chrismytton)

tmtmtmtm:
superseded by new approach

PR #14663, opened 2016-08-17
Extract SpiderBase class

ondenman:
Move logic from `/status/all_countries.html` to a Page class

Test page: https://everypolitician-viewe-pr-14663.herokuapp.com/status/all_countries.html

### DIFF
--- a/lib/page/all_countries.rb
+++ b/lib/page/all_countries.rb
@@ -1,0 +1,10 @@
+require_relative '../world.rb'
+
+module Page
+  class AllCountries
+    attr_accessor :world
+    def initialize
+      @world = World.new.as_json.to_a
+    end
+  end
+end
--- a/lib/page/all_countries.rb
+++ b/lib/page/all_countries.rb
@@ -4,7 +4,7 @@ module Page
   class AllCountries
     attr_accessor :world
     def initialize
-      @world = World.new.as_json.to_a
+      @world = World.new.as_json
     end
   end
 end
--- a/lib/page/all_countries.rb
+++ b/lib/page/all_countries.rb
@@ -2,9 +2,8 @@ require_relative '../world.rb'
 
 module Page
   class AllCountries
-    attr_accessor :world
-    def initialize
-      @world = World.new.as_json
+    def world
+      World.new.as_json
     end
   end
 end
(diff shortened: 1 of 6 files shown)


t/page/all_countries.rb
```diff
@@ -0,0 +1,20 @@
...
+  subject do
+    Page::AllCountries.new
+  end
+
+  it 'should return an array' do
+    subject.world.must_be_instance_of Array
+  end
+
+  it 'should return a list of countries' do
+    countries = subject.world.flatten
```
tmtmtmtm: Why do you need to `flatten` this? If the actual structure coming back here is complex enough to need that, we might need a test for the real structure.

t/page/all_countries.rb
```diff
@@ -0,0 +1,20 @@
...
+    a = countries.each.include? :argentina
+    b = countries.each.include? :belgium
+    c = countries.each.include? :china
```
tmtmtmtm: What is the scenario where a and b are true, but c is false?

lib/page/all_countries.rb
```diff
@@ -0,0 +1,10 @@
+require_relative '../world.rb'
+
+module Page
+  class AllCountries
+    attr_accessor :world
+    def initialize
+      @world = World.new.as_json.to_a
```
tmtmtmtm: Why are you converting this to an Array? The underlying structure is Hash-like, so seems like it would simpler to use that rather than flattening to a very odd interleaved array.

t/page/all_countries.rb
```diff
@@ -0,0 +1,20 @@
...
+    Page::AllCountries.new
+  end
+
+  it 'should return an array' do
+    subject.world.must_be_instance_of Array
+  end
+
+  it 'should return a list of countries' do
+    countries = subject.world.flatten
+    a = countries.each.include? :argentina
```
tmtmtmtm: Why the `.each` here?

t/page/all_countries.rb
```diff
@@ -0,0 +1,20 @@
...
+    b = countries.each.include? :belgium
```
tmtmtmtm: Seems a little odd that the expectation is `:belgium` rather than "Belgium".

t/page/all_countries.rb
```diff
@@ -0,0 +1,20 @@
...
+    c = countries.each.include? :china
+    (a & b & c).must_equal true
```
tmtmtmtm: This is cute, but I think it would be much clearer to simply have `countries.must_include 'Belgium'` (and repeated similar tests if required, though it's not clear why we'd need those)

PR #14678, opened 2016-08-17
Add minitest to the Gemfile

octopusinvitro:
If minitest is not installed globally in the system, running the tests will throw an error 'Can't find minitest/autorun'.

Adding the gem and bundle installing it locally solves the issue.


tmtmtmtm:
Possibly worth explicitly noting in the commit that this is needed because minitest was removed from core in ruby 2.2.0, otherwise LGTM 👍

PR #14682, opened 2016-08-17
Extract a countries page class

octopusinvitro:
This PR extracts a countries page class from the app route.

### DIFF
--- a/lib/page/countries.rb
+++ b/lib/page/countries.rb
@@ -1,0 +1,18 @@
+require 'everypolitician'
+require_relative '../world'
+
+module Page
+  class Countries
+    def countries
+      @countries ||= EveryPolitician.countries
+    end
+
+    def download_url
+      @download_url ||= EveryPolitician.countries_json
+    end
+
+    def world
+      @world ||= World.new.as_json
+    end
+  end
+end
--- a/lib/page/countries.rb
+++ b/lib/page/countries.rb
@@ -3,16 +3,24 @@ require_relative '../world'
 
 module Page
   class Countries
+    def initialize(index:)
+      @index = index
+    end
+
     def countries
-      @countries ||= EveryPolitician.countries
+      index.countries
     end
 
     def download_url
-      @download_url ||= EveryPolitician.countries_json
+      index.index_url
     end
 
     def world
       @world ||= World.new.as_json
     end
+
+    private
+
+    attr_reader :index
   end
 end
--- a/lib/page/countries.rb
+++ b/lib/page/countries.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'everypolitician'
 require_relative '../world'
 
--- a/t/page/countries.rb
+++ b/t/page/countries.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'minitest/autorun'
 require_relative '../../lib/page/countries'
 require 'pry'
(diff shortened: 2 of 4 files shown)


lib/page/countries.rb
```diff
@@ -0,0 +1,18 @@
...
+    def countries
+      @countries ||= EveryPolitician.countries
+    end
+
+    def download_url
+      @download_url ||= EveryPolitician.countries_json
+    end
+
+    def world
+      @world ||= World.new.as_json.to_a
```
tmtmtmtm: `World.new.as_json` returns a Hash, so it's a little odd to coerce it to an Array. The only thing that the template does with this at the moment is call `.length` on it, which gets the right thing either way, but if the template tried to do anything extra with this it could get quite confusing. 

We'll probably want a follow-up PR to remove that logic entirely from the template, and only provide a `missing_country_count` method here instead, but until we do that I think I'd be a bit more comfortable if we just passed the Hash around here instead.

t/page/countries.rb
```diff
@@ -0,0 +1,26 @@
...
+CJSON = 'https://cdn.rawgit.com/everypolitician/everypolitician-data/%s/countries.json'.freeze
+SHA   = 'd8a4682f'.freeze
+
+describe 'Countries' do
+  subject do
+    Page::Countries.new
+  end
+
+  it 'has a list of all countries in Everypolitician' do
+    assert_equal subject.countries.first[:name], 'Abkhazia'
```
tmtmtmtm: All the other test files use the ".must_equal" syntax rather than the `assert_equal` syntax. I don't massively care either way which we use, but we should be consistent.

t/page/countries.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  it 'has a url pointing to the last version of the db' do
+    datasource = CJSON % SHA
+    Everypolitician.countries_json = datasource
```
tmtmtmtm: I'm not sure that setting this at this point has the effect you think it's having. And, worse, I'm not sure how to _be_ sure that it does. The logic here might be simpler to follow if *all* the tests were using the same datasource, and this one was just testing that it was the expected one.

t/page/countries.rb
tmtmtmtm: Now that this is a Country object from `everypolitician-ruby`, I think it probably would be better to test `.name` rather than `:name`, yes (and probably do likewise in the template).

lib/page/countries.rb
octopusinvitro: Ah, I had the same thought, but I just copied what was in the route. Shame on me!

t/page/countries.rb
```diff
@@ -2,25 +2,27 @@
...
+datasource = CJSON % SHA
 
 describe 'Countries' do
   subject do
     Page::Countries.new
   end
 
   it 'has a list of all countries in Everypolitician' do
-    assert_equal subject.countries.first[:name], 'Abkhazia'
+    Everypolitician.countries_json = datasource
```
tmtmtmtm: Ah, when I said before to make all tests use the same setup, I meant in a `before`:

```ruby
  before { Everypolitician.countries_json = CJSON % SHA }
```

tmtmtmtm:
Grrr. These tests pass locally for me — and https://everypolitician-viewe-pr-14682.herokuapp.com/countries.html looks fine too…

@chrismytton can you spot anything obvious here?

PR #14708, opened 2016-08-18
separate release and build process

struan:
Split the viewer-sinatra build and deploy process into two functions and always call the build process. The deploy process is only called when we are on the master branch.


chrismytton:
@struan Looks like Travis was failing because it couldn't find the deploy key. I've pushed an extra commit to add the deploy key before doing the build.

Other than that this all looks good to me! 👍

PR #14716, opened 2016-08-18
Switch to Everypolitician::Index

chrismytton:
This switches the existing page classes and their tests over to use `Everypolitician::Index`. It's currently pointing at the version of everypolitician-ruby in https://github.com/everypolitician/everypolitician-ruby/pull/44, so that will need to be changed back to `master` once that PR is merged.

### DIFF
--- a/lib/page/download.rb
+++ b/lib/page/download.rb
@@ -4,14 +4,16 @@ require 'json'
 module Page
   class Download
     attr_reader :download_url
+    attr_reader :index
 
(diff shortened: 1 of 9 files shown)


lib/page/download.rb
```diff
@@ -4,14 +4,16 @@
 module Page
   class Download
     attr_reader :download_url
+    attr_reader :index
```
tmtmtmtm: I think this should probably private

PR #14736, opened 2016-08-19
Add "frozen_string_literal: true" everywhere

tmtmtmtm:
Now that we're on Ruby 2.3 Rubocop is suggesting that we use the new
'frozen_string_literal' everywhere, in preparation for 3.0. Helpfully,
`rubocop -a` even does it for us…

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,8 @@
 AllCops:
-  Exclude: 
+  Exclude:
     - 'Vagrantfile'
     - 'vendor/**/*'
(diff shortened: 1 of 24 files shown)


tmtmtmtm:
@chrismytton any idea why rubocop is failing here, but passing locally for me?

chrismytton:
Looks like this was passing locally but not on Travis because we've got `.ruby-version` files locally but there isn't one on Travis, so it wasn't detecting the Ruby version correctly. https://github.com/bbatsov/rubocop/issues/2956

PR #14737, opened 2016-08-19
named parameters everywhere

tmtmtmtm:
Now that we're on ruby 2.3.1 we can be stricter about always requiring
named parameters to Page classes, and more flexible about how we say which
ones are required.


chrismytton:
Looks good to me! 👍

PR #14741, opened 2016-08-19
Stub out request for known countries.json in tests

chrismytton:
This change stubs out the http requests we're making in the tests so that they run fast and don't depend on an internet connection.


chrismytton:
OK, for now this is just the quick win version that stubs out the known `countries.json` request.

PR #14743, opened 2016-08-19
Add a `missing_countries` method in Page::Countries

tmtmtmtm:
Remove logic from the template where it doesn't belong, and pull it back up
into the Page class.


tmtmtmtm:
https://everypolitician-viewe-pr-14743.herokuapp.com/countries.html still looks right:

screen shot 2016-08-19 at 15 22 43

PR #14749, opened 2016-08-19
Use VCR to record http requests in the tests

chrismytton:
This change adds VCR and makes it so that any http interactions in the tests are *automatically* recorded and given an appropriate name. As part of this I've tidied up the duplication between the tests and pulled it all into `test_helper.rb` so that all tests run in the same environment.


tmtmtmtm:
I've cherry-picked some of more drive-by commits out as separate PRs: 
* https://github.com/everypolitician/viewer-sinatra/pull/14750
* https://github.com/everypolitician/viewer-sinatra/pull/14751

There's only one route left that needs ALL_COUNTRIES, so I've also inlined that there (https://github.com/everypolitician/viewer-sinatra/pull/14753), which should let you simplify https://github.com/everypolitician/viewer-sinatra/pull/14749/commits/8219fff60eefbbf31572b4ce1da0508f91c2bead

tmtmtmtm:
LGTM (after a fixup to keep rubocop happy)

PR #14845, opened 2016-08-22
Fix header link

### LINKED ISSUE #14816: 'Country' link in header is broken
On pages like http://everypolitician.org/estonia/riigikogu/term-table/13.html, the 'Estonia' link in the header goes to '//' rather than '/estonia/'

octopusinvitro:
The `:url` member of the `Country` struct was renamed to `:slug` to play well with the real `Country` objects.

The main layout was calling a `:url` field in the `@country` returned by the currently displayed page, but `Country` objects don't have a url member, they have a `:slug` instead.

So by making the titles work for world countries, we broke it for EP countries.
After renaming `:url` to `:slug` in world countries, it should work for both types of country.

The main layout headings now call `@country[:slug]` instead of `@country[:url]`
so that both world-countries and EP-countries heading link to the right url.

Fixes #14816

### DIFF
--- a/lib/world.rb
+++ b/lib/world.rb
@@ -15,7 +15,7 @@ class World
   # want to tidy the interface up a little so that this is substitutable
   # for an `everypolitician-ruby` Country in a few well-defined places.
 
-  Country = Struct.new(:url, :name, :names)
+  Country = Struct.new(:slug, :name, :names)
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
👍

PR #14846, opened 2016-08-22
Use country methods in template

octopusinvitro:
Instead of calling keys on the hash, the template now uses the methods provided by the `Country` object.


tmtmtmtm:
👍

PR #14847, opened 2016-08-22
Fix the spider_base template

octopusinvitro:
The `spider_base` template now uses the `countries` method instead of manipulating a JSON object.

### DIFF
--- a/lib/page/spider_base.rb
+++ b/lib/page/spider_base.rb
@@ -3,12 +3,18 @@ require_relative '../world.rb'
 
 module Page
   class SpiderBase
-    def world
-      World.new.as_json
-    end
-
     def title
       'EveryPolitician: Robots Start Here'
+    end
+
+    def countries
+      world.countries
+    end
+
+    private
+
+    def world
+      World.new
     end
   end
 end
--- a/lib/world.rb
+++ b/lib/world.rb
@@ -10,6 +10,10 @@ class World
     @_wjson ||= JSON.parse(File.read(file), symbolize_names: true)
   end
 
+  def countries
+    as_json.keys.map { |slug| country(slug) }.sort_by(&:name)
+  end
+
   # super-simplistic adapter for the inner data. Over time it might make
   # sense to properly extract this to its own class, but for now we just
   # want to tidy the interface up a little so that this is substitutable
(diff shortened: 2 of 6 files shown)


views/all_countries.erb
```diff
@@ -1,11 +1,13 @@
 <div class="page-section" id="home">
     <div class="container">
         <ul class="grid-list grid-list--vertically-center">
-          <% @page.world.sort_by { |k, _| k }.each do |k, v| %>
+          <% @page.world.as_json.sort_by { |key, _| key }.each do |key, value| %>
```
tmtmtmtm: The goal here is to kill that `as_json` method. You'll probably want to pass `world.all_countries` or equivalent into the template which provides a list of country objects.

lib/world.rb
```diff
@@ -10,6 +10,10 @@ def as_json
     @_wjson ||= JSON.parse(File.read(file), symbolize_names: true)
   end
 
+  def countries
+    as_json.map { |slug, _| country(slug) }
```
tmtmtmtm: Would probably be a little clearer to do that as `as_json.keys.map { |slug| country_slug }`

tmtmtmtm: I'd probably also have this guarantee to return the countries already sorted (by name). Most uses are probably going to want that, so may as well save them the work.

lib/page/spider_base.rb
```diff
@@ -3,12 +3,18 @@
...
-    def world
-      World.new.as_json
-    end
-
     def title
       'EveryPolitician: Robots Start Here'
     end
+
+    def countries_sorted
+      world.countries.sort_by(&:slug)
```
tmtmtmtm: Sorting by slug is the existing behaviour, but as this is essentially a private page, I don't think it's essential to maintain that order. So if `world.countries` were to return a list ordered by name, I think it would be OK to pass it straight through without resorting here.

t/web/spider_base.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../app'
+require 'pry'
```
tmtmtmtm: I don't think you want to leave that behind. (Though might be worth making a separate PR to just add it to `test_helper` so that all tests have it available)

octopusinvitro:
@tony I rebased the branchso tidy things up a bit.

tmtmtmtm:
LGTM!

PR #14849, opened 2016-08-22
Do not pass popolo data to wikidata page

### LINKED ISSUE #14839: Don't pass Popolo to Wikidata page
Generally we shouldn't be passing the entire Popolo object into templates, but only the parts needed for the page. 

Clean this up for `house_wikidata`, passing lists of `people_with_wikidata` and `people_without_wikidata` instead.

ondenman:
HouseWikidata now passes two arrays. One is a list of people with wikidata, the other is a list of people without.

Closes #14839


lib/page/house_wikidata.rb
```diff
@@ -12,21 +13,49 @@ def title
...
 
+    def people_with_wikidata
+      result = popolo.persons.reduce([]) do |arr, person|
+        if scheme = person.identifiers.select { |p| p[:scheme] == 'wikidata' }
+          wikidata_identifier = scheme.first[:identifier]
+          wikidata_url = "https://www.wikidata.org/wiki/#{wikidata_identifier}"
+          arr.push( { name: person.name,
+                      wikidata_identifier: wikidata_identifier,
+                      wikidata_url: wikidata_url
+                    }
```
tmtmtmtm: I don't think we need to collapse this down to our own data structure here — it's fine just to pass in a list of `EveryPolitician::Person` objects

lib/page/house_wikidata.rb
```diff
@@ -12,21 +13,49 @@ def title
...
+                  )
+        end
+      end
+      return result || []
+    end
+
+    def people_without_wikidata
+      result = popolo.persons.reduce([]) do |arr, person|
+        unless person.identifiers.select { |p| p[:scheme] == 'wikidata' }
+          arr.push({ name: person.name })
```
tmtmtmtm: Same as above: passing `person` is fine here.

lib/page/house_wikidata.rb
```diff
@@ -12,21 +13,49 @@ def title
       "EveryPolitician: #{country.name} — #{house.name}"
     end
 
-    def country
-      index.country(country_slug)
+    def house_name
+      house.name
```
tmtmtmtm: Not quite sure what this is for, but https://github.com/everypolitician/viewer-sinatra/issues/14816 might have solved the underlying problem already.

lib/page/house_wikidata.rb
```diff
@@ -12,21 +13,49 @@ def title
...
     end
 
-    def house
-      country.legislature(house_slug)
+    def country
+      index.country(country_slug)
     end
 
     # TODO: we shouldn't be passing raw Popolo, only what's needed
     def popolo
```
tmtmtmtm: One of the key goals here is to remove this method (or move it to private)

lib/page/house_wikidata.rb
```diff
@@ -12,21 +13,49 @@ def title
...
       house.popolo
     end
 
+    def people_with_wikidata
```
tmtmtmtm: This can all be massively simplified by using `partition`, as the existing template already does. It will divide an array into two separate arrays, based on some condition (e.g. whether the entries have wikidata or not!)

t/page/house_wikidata.rb
```diff
@@ -11,16 +11,20 @@
...
   end
 
   it 'should return a title' do
     subject.title.must_equal 'EveryPolitician: Austria — Nationalrat'
   end
 
-  it 'should have popolo with wikidata' do
-    andrea = '0eedf2c9-01ea-44f4-bc6e-e5e4bf6d2add'
-    subject.popolo.persons.find_by(id: andrea).wikidata.must_equal 'Q493950'
+  it 'should pass a list of people with wikidat' do
```
tmtmtmtm: typo: "wikidat"

lib/page/house_wikidata.rb
```diff
@@ -12,21 +13,33 @@ def title
...
 
     attr_reader :house_slug, :country_slug, :index
+
+    def popolo
+      house.popolo
+    end
+
+    def people_with_and_without_wikidata
+      @with_and_without_array ||= popolo.persons.partition(&:wikidata)
+      { with: @with_and_without_array[0], without: @with_and_without_array[1] }
```
tmtmtmtm: it seems a little needless to convert this to a hash here, rather than just calling `people_with_and_without_wikidata.first` or `.last` in the public methods.

lib/page/house_wikidata.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require_relative '../popolo_helper'
```
tmtmtmtm: why did you need to add this? I can't see anything that should require it, and removing it doesn't cause any obvious failure.

t/page/house_wikidata.rb
```diff
@@ -19,8 +19,21 @@
     subject.title.must_equal 'EveryPolitician: Austria — Nationalrat'
   end
 
-  it 'should have popolo with wikidata' do
-    andrea = '0eedf2c9-01ea-44f4-bc6e-e5e4bf6d2add'
-    subject.popolo.persons.find_by(id: andrea).wikidata.must_equal 'Q493950'
+  it 'should pass a list of people with wikidata' do
+    res = subject.people_with_wikidata.select { |p| p[:name] == 'Cornelia Ecker' }
```
tmtmtmtm: Straggling `[:name]` rather than `.name`. This might be clearer, though, as 

```ruby
subject.people_with_wikidata.map(&:name).must_include 'Cornelia Ecker'
```

t/page/house_wikidata.rb
```diff
@@ -19,8 +19,21 @@
...
+    res.wont_be_nil
+  end
+
+  it 'should pass a list of people without wikidata as an array' do
+    subject.people_without_wikidata.must_be_empty
```
tmtmtmtm: It's a little worrying to have _only_ a test that this list is empty. This would still pass even if `people_without_wikidata` always returned `[]`. You probably want to add a test for another country too, where at least a few people don't have Wikidata mappings.

tmtmtmtm:
Yay! 👍  

There are a lot of VCR cassettes here that I don't think are needed, but I want to squash these commits down a bit first anyway, so I'll clean those out whilst I'm doing that.

PR #14891, opened 2016-08-23
14844 remove json from country selector

ondenman:
country-selector now works with Country struct provided by world.rb
Have added `:total_people` accessor to Country struct
`home.rb#world` now returns struct
Edited .erb accordingly.

### DIFF
--- a/lib/page/home.rb
+++ b/lib/page/home.rb
@@ -25,8 +25,8 @@ module Page
     end
 
     def world
-      world_json.each do |slug, country|
-        country[:totalPeople] = index.country(slug.to_s).person_count rescue 0
+      world_countries.each do |country|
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
This approach is quite nasty. You're adding a field to a Class (albeit as a Struct) that only you know about it, and then filling it in yourself. But any other user of World won't know anything about that — they'll just get a blank field there.

In general the best way to solve things like this is to decorate the object you want to add something to. But here I think we need to tie together `World` and `everypolitician-country` better (not least because we'll be able to use that directly in lots of other places too). I've made a first stab at that in https://github.com/everypolitician/viewer-sinatra/pull/15066. It's a bit ugly for now, but we'll be able to clean that up separately later.

PR #14903, opened 2016-08-23
Remove all brackets from the term page class

octopusinvitro:
When I started working on another template I realized that the term table class was using a lot of brackets as well instead of the methods, so I decided to make this change first before updating the term table template.

Things spotted but not fixed here, as they will require a bit more work, so maybe they must each have their own PR:

* in the Popolo::JSON instance, we are calling `popolo.popolo[:meta][:sources]`. Maybe we could add a method to that class to return these sources.

* in `current_term[:csv_url]`, this is something that was proposed but not added yet, there is an issue about it: https://github.com/everypolitician/everypolitician-ruby/issues/32 

* in `house[:popolo]` I guess it refers to the `popolo_url`, however, I think a new vcr cassete might be needed, as this happens at the moment:

```ruby
> house[:popolo]
"data/Northern_Ireland/Assembly/ep-popolo-v1.0.json"
> house.popolo_url
"https://cdn.rawgit.com/everypolitician/everypolitician-data/1a17862/data/Northern_Ireland/Assembly/ep-popolo-v1.0.json"
```
so the tests fail if I change that. Will make a specific PR for that.

### DIFF
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -12,7 +12,7 @@ module Page
     end
 
     def title
-      parts = [country.name, house[:name], current_term[:name]]
+      parts = [country.name, house.name, current_term.name]
       "EveryPolitician: #{parts.join(' - ')}"
     end
 
@@ -25,13 +25,13 @@ module Page
     end
 
     def house
-      country[:legislatures].find do |h|
-        h[:slug].downcase == house_slug.downcase
+      country.legislatures.find do |house|
+        house.slug.downcase == house_slug.downcase
       end
     end
 
     def terms
-      house[:legislative_periods]
+      house.legislative_periods
     end
 
     def next_term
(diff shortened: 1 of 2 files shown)


lib/page/term_table.rb
```diff
@@ -25,13 +25,13 @@ def country
     end
 
     def house
-      country[:legislatures].find do |h|
-        h[:slug].downcase == house_slug.downcase
+      country.legislatures.find do |house|
+        house.slug.downcase == house_slug.downcase
```
tmtmtmtm: I think this should just be `country.legislatures(house_slug)` now

lib/page/term_table.rb
```diff
@@ -139,7 +139,7 @@ def popolo
     end
 
     def popolo_file
-      @popolo_file ||= EveryPolitician::GithubFile.new(house[:popolo], house[:sha])
+      @popolo_file ||= EveryPolitician::GithubFile.new(house[:popolo], house.sha)
```
tmtmtmtm: Do some of the other TODOs mean that we can't just go directly to `house.popolo` here yet?

octopusinvitro:
@tony I will address your two comments in future PR, as a lot of things broke when I made the changes. :+1:

PR #14908, opened 2016-08-23
Replace references by key with method calls

ondenman:
Needed template now works with objects

### DIFF
--- a/lib/page/needed.rb
+++ b/lib/page/needed.rb
@@ -12,15 +12,15 @@ module Page
     end
 
     def to_find
-      (issues 'New Country,To Find').sort_by(&:title)
+      issues 'New Country,To Find'
     end
 
     def to_scrape
-      (issues 'New Country,To Scrape').sort_by(&:title)
+      issues 'New Country,To Scrape'
     end
 
     def to_finish
-      (issues 'New Country,3 - WIP').sort_by(&:title)
+      issues 'New Country,3 - WIP'
     end
 
     private
@@ -28,7 +28,7 @@ module Page
     attr_reader :access_token
 
     def issues(labels)
-      client.issues 'everypolitician/everypolitician-data', labels: labels
+      (client.issues 'everypolitician/everypolitician-data', labels: labels).sort_by(&:title)
     end
 
     def client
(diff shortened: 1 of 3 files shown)


views/needed.erb
```diff
@@ -25,11 +25,11 @@
 
         <div class="container">
             <ul class="grid-list grid-list--vertically-center source-needed">
-              <% @page.to_find.sort_by { |i| i[:title] }.each do |i| %>
+              <% @page.to_find.sort_by { |i| i.title }.each do |i| %>
```
tmtmtmtm: These could all be collapsed now to `sort_by(&:title)`, but the deeper issue is that really it shouldn't be up to the _template_ to do that. As we're doing that in all three cases, it would really make much more sense to push it back into the Page class, and have it guarantee to give them in sorted order. 

Up to you whether to make a new Issue for that, and reference it, or do it within this PR…

lib/page/needed.rb
```diff
@@ -12,15 +12,15 @@ def title
     end
 
     def to_find
-      issues 'New Country,To Find'
+      (issues 'New Country,To Find').sort_by(&:title)
```
tmtmtmtm: I think it would be much cleaner to add this to the `issues` method, rather than every method that calls that.

t/page/needed.rb
```diff
@@ -9,7 +9,12 @@
   end
 
   it 'should know that Eritrea is missing' do
-    subject.to_find.map(&:title).include? 'Eritrea'
+    assert subject.to_find.map(&:title).include? 'Eritrea'
+  end
+
+  it 'should order the list of countries to find by name' do
+    order = subject.to_find.map(&:title).first <=> subject.to_find.map(&:title).last
+    assert order <= 0
```
tmtmtmtm: This is very complex logic to have in a test (and tests that effectively say "foo should be either X or Y" are a also a red flag — here you're saying it should be zero _or_ less than")

One way to express the intent more clearly might be something like:
```ruby
wanted = subject.to_find.map(&:title)
assert wanted == wanted.sort
```

tmtmtmtm:
https://github.com/everypolitician/viewer-sinatra/pull/14908/commits/083f1ed6aaaed08598c2b48aebe40d051416b105 should probably have been its own PR, so we could have had the benefit from it straightaway, but this LGTM now! 👍  :shipit:

PR #14948, opened 2016-08-24
Add house.csv_url to download template

ondenman:
Adds `house.csv_url` to template. Removes house[:names].

### DIFF
--- a/t/web/download.rb
+++ b/t/web/download.rb
@@ -5,9 +5,19 @@ require_relative '../../app'
 describe 'Download names.csv button' do
   subject { Nokogiri::HTML(last_response.body) }
   before { get '/united-states-of-america/download.html' }
-  it 'should link to the names.csv file' do
-    subject_buttons = subject.css('a.button--download/@href').map(&:value).select { |h| h.include? '/names.csv' }
-    subject_buttons.must_include 'https://cdn.rawgit.com/everypolitician/everypolitician-data/735c68a/data/United_States_of_America/House/names.csv'
-    subject_buttons.must_include 'https://cdn.rawgit.com/everypolitician/everypolitician-data/cb4a755/data/United_States_of_America/Senate/names.csv'
+
+  let(:links) do
+    subject.css('a.button--download/@href')
+           .map(&:value)
+           .select { |h| h.include? '/names.csv' }
+           .join
+  end
+
+  it 'should link to the House names.csv file' do
+    links.must_include '735c68a/data/United_States_of_America/House/names.csv'
+  end
+
+  it 'should link to the Senate names.csv file' do
+    links.must_include 'cb4a755/data/United_States_of_America/Senate/names.csv'
   end
 end
--- a/t/web/download.rb
+++ b/t/web/download.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../app'
+
+describe 'Download names.csv button' do
+  subject { Nokogiri::HTML(last_response.body) }
+  before { get '/united-states-of-america/download.html' }
+  it 'should link to the names.csv file' do
+    subject_buttons = subject.css('a.button--download/@href').map(&:value).select { |h| h.include? '/names.csv' }
+    subject_buttons.must_include 'https://cdn.rawgit.com/everypolitician/everypolitician-data/735c68a/data/United_States_of_America/House/names.csv'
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
In principle this looks good, but I'm a little scared that there doesn't seem to be a test that covers this. Probably worth taking this opportunity to add something super-simple just to check that these URLs look correct for a known country (preferably a bicameral one)

tmtmtmtm:
I think there are couple of different issues wrapped up together here, that require some untangling. Firstly, I think the `csv_url` method that this is using was badly named, and making it hard to see what was going on here. I've fixed that in https://github.com/everypolitician/everypolitician-ruby/pull/57

On top of that, the test that you've added here is at the wrong level. The change you've made is to an `.erb` file, and Page classes (and tests) are entirely unaware of those. The tests you've introduced here are only testing things that are outside of the scope of the Page class being tested — I can't think of anything you could change in `Page::Download` that would make this test fail that wouldn't already be covered by existing tests.

Changes to a template generally require a `t/web/` level integration test, instead.

tmtmtmtm:
Not sure what the Travis failure here is all about. The (well hidden) errors are:

```
2016-08-25 11:50:57 URL:http://localhost:4567/peru/download.html [15779/15779] -> "localhost:4567/peru/download.html" [1]
2016-08-25 11:50:58 +0000: Rack app error: #<OpenURI::HTTPError: 404 Not Found>
…
http://localhost:4567/peru/congreso/term-table/2011.html:
2016-08-25 11:50:58 ERROR 500: Internal Server Error.```
```

But the other run seemed fine, and I can't reproduce this locally.

t/web/download.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../app'
+
+describe 'Viewer' do
+  subject { Nokogiri::HTML(last_response.body) }
+
+  describe 'when viewing the home page' do
```
tmtmtmtm: This isn't really about viewing the home page…

t/web/download.rb
```diff
@@ -0,0 +1,18 @@
...
+    before { get '/united-states-of-america/download.html' }
+
+    it 'should have show some text' do
```
tmtmtmtm: This doesn't seem like the correct description of this test either.

t/web/download.rb
```diff
@@ -0,0 +1,18 @@
...
+      last_response.body.must_include 'https://cdn.rawgit.com/everypolitician/everypolitician-data/'
```
tmtmtmtm: This will find this anywhere in the body, e.g. on the "Download CSV" links for the term data.

t/web/download.rb
```diff
@@ -0,0 +1,18 @@
...
+      last_response.body.must_include '/data/United_States_of_America/'
```
tmtmtmtm: ditto for this.

t/web/download.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../app'
+
+describe 'Viewer' do
+  subject { Nokogiri::HTML(last_response.body) }
```
tmtmtmtm: You're setting up a 'subject' here that you don't seem to ever use.

t/web/download.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../app'
+
+describe 'Viewer' do
```
tmtmtmtm: The context here isn't really about a 'Viewer'

t/web/download.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  describe 'when viewing the home page' do
+    before { get '/united-states-of-america/download.html' }
+
+    it 'should have show some text' do
+      last_response.body.must_include 'https://cdn.rawgit.com/everypolitician/everypolitician-data/'
+      last_response.body.must_include '/data/United_States_of_America/'
+      last_response.body.must_include 'House/names.csv'
+      last_response.body.must_include 'Senate/names.csv'
+    end
```
tmtmtmtm: I suspect what you actually want to be testing here is something `subject.css('a.button--download/@href').map(&:value).select { |h| h.include? '/names.csv' }`

tmtmtmtm:
I've rebased this over `master` to hopefully fix the test that's failing due to the sha collision. In the process I dropped the commits to add and remove the needless test, and squashed a few of the others together.

tmtmtmtm:
I've also tidied the test up to be a little more rubocop friendly, so this looks Good To Go.

PR #14952, opened 2016-08-24
Remove csv_url method from Page::TermTable

octopusinvitro:
<s>Now that the EP-ruby has been updated to use the right `csv_url` coming from the data in terms, we can switch to calling the method instead of accessing the raw data through the hash key.</s>

The Term Table page no longer displays download links directly (these are now on the Download page), so we don't need a  `csv_url` method.

### DIFF
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -47,7 +47,7 @@ module Page
     end
 
     def csv_url
-      current_term[:csv_url]
+      current_term.csv_url
     end
 
     def popolo_url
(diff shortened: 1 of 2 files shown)


chrismytton:
I don't think `csv_url` is actually being used anywhere in `views/term_table.erb`. Perhaps it would be better to just  remove this method?

tmtmtmtm:
Yep, looks to me like the only place using this at all is the download page.

tmtmtmtm:
The description here doesn't seem to match what's actually being done in the PR. I'm guessing it just hasn't been updated after https://github.com/everypolitician/viewer-sinatra/pull/14952/commits/aaaa6062809707757a2d89793fdcd39fd93f237e changed things?

PR #14954, opened 2016-08-24
Remove githubfile class

octopusinvitro:
There is no need to use the `GitHubFile` helper anymore to build the popolo url and file, since legislatures have a `popolo` and `popolo_url` methods that can be used instead.

Also, since this class is not needed anymore, as EP-ruby and EP-popolo libraries already provide that functionality, this class is not used anywhere at this point.

This PR removes that class from the `CountryExtensions` helper and uses the proper methods from the EP-ruby legislature class.

### DIFF
--- a/lib/everypolitician_extensions.rb
+++ b/lib/everypolitician_extensions.rb
@@ -9,28 +9,6 @@ module EveryPolitician
       legislatures.map(&:person_count).inject(:+)
     end
   end
-
-  class GithubFile
-    GH_PATH = 'https://cdn.rawgit.com/everypolitician/everypolitician-data/%s/%s'
-
-    # TODO: investigate whether we can remove this caching
-    # (possibly easier to just remove this entire class)
-    def initialize(file, sha, cache_dir = '_cached_data')
-      @url = GH_PATH % [sha, file]
-
-      FileUtils.mkpath cache_dir
-      @cache_file = File.join cache_dir, sha + '-' + file.tr('/', '-')
-    end
-
-    attr_reader :url
-
-    def raw
-      @_data ||= begin
-        File.write(@cache_file, open(@url).read) unless File.exist? @cache_file
-        File.read(@cache_file)
-      end
-    end
-  end
 end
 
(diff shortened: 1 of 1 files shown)


lib/page/term_table.rb
```diff
@@ -49,7 +49,7 @@ def csv_url
     end
 
     def popolo_url
-      popolo_file.url
+      house.popolo_url
     end
```
chrismytton: Looks like this method isn't being used in the term table view anymore, so probably best to just remove it.

tmtmtmtm:
This PR feels a little bit too like a ragbag of only semi-related things. I would expect a PR that declares itself to be "Get rid of X class" to have nothing else in it _other_ than removing the class (and related tests etc). The changes that _enable_ getting rid of it seem like they belong in a PR that more accurately describes what *they* are doing.

tmtmtmtm:
cherry picked https://github.com/everypolitician/viewer-sinatra/pull/14954/commits/f9cf7ffc3f2a97b8e67e977d4a8fe50986f97d36 as https://github.com/everypolitician/viewer-sinatra/pull/15051

lib/page/term_table.rb
tmtmtmtm: Done in https://github.com/everypolitician/viewer-sinatra/pull/15053

tmtmtmtm:
https://github.com/everypolitician/viewer-sinatra/pull/14954/commits/df00d17f3751d927e2b1001c72fdfd93afcd69cf implemented in https://github.com/everypolitician/viewer-sinatra/pull/15053 (a cherry-pick didn't apply cleanly, and the version here didn't have any commit message, so done as a new commit)

tmtmtmtm:
https://github.com/everypolitician/viewer-sinatra/pull/14954/commits/ac974dcfc979a7e7a561b5b710ad761095ef3c05 cherry-picked as https://github.com/everypolitician/viewer-sinatra/pull/15059

tmtmtmtm:
Now that this is only about removing the file, I've rebased out the commits we cherry-picked to elsewhere.

PR #14972, opened 2016-08-24
Use popolo persons - WIP

octopusinvitro:
The `people` method now returns a collection of `Popolo::Person` objects. 

While we waited for #14954 to be merged, we decided to temporarily use `house.popolo` and so avoid using the `popolo_file` method inside the `people` method, since it uses the `GithubFile` helper that we want to get rid of.

The template was receiving every person as a hash, but the hash returned by `people` in the term table page was adding information to the person that is not present in a `Popolo::Person` instance.

 In particular, the extra information added to a person hash is:
* `:social`,
* `:bio`,
* `:contacts` and 
* `:identifiers`,

This extra information is then used to build the `percentages`.

We think those methods don't belong in the `Popolo::Person` instances, so in order to break as less things as possible, we decided to just move those methods to a helper module for now, and use them as if they were methods of the class, with the intention to remove them from there in the future, so that we could clean the `people` method.

So for these four hash keys in the person hash, we created temporary methods in the `PersonExtension` module, (in everypolitician_extensions.rb) which we then included in the `Popolo::Person` instance (line 110 in everypolitician_extensions):

```ruby
EveryPolitician::Popolo::Person.include EveryPolitician::PersonExtension
```

We had some problems with the proxy image, so it is commented for now.

The person memberships brackets were not removed yet in the template.

We temporarily commented the tests that are failing so that we can solve them one at a time (last commit). This is where we are now.

### DIFF
--- a/lib/everypolitician_extensions.rb
+++ b/lib/everypolitician_extensions.rb
@@ -31,6 +31,80 @@ module EveryPolitician
       end
     end
   end
+
+  module PersonExtension
+    def social
+      social = []
+      if twitter
+        social << {
+          type:  'Twitter',
+          value: "@#{twitter}",
+          link:  "https://twitter.com/#{twitter}",
+        }
+      end
+
+      if facebook
+        fb_username = URI.decode_www_form_component(facebook.split('/').last)
+        social << {
+          type:  'Facebook',
+          value: fb_username,
+          link:  "https://facebook.com/#{fb_username}",
+        }
+      end
+      social
+    end
+
+    def bio
+      bio = []
+      bio << { type: 'Gender', value: gender } if gender
+      bio << { type: 'Born', value: birth_date } if birth_date
+      bio << { type: 'Died', value: death_date } if death_date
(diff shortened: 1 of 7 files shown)


tmtmtmtm:
I'm not convinced that this is the right approach to this. The data structures in question here are groupings that are very specific to the display in a particular context — as you say, they're not really general enough structures about the Person to be considered for pushing up into `everypolitician-popolo`. But that means that they shouldn't really be patched locally onto `EveryPolitician::Person` either. 

I haven't thought too deeply about this yet, but my initial guess is that we should have Person Cards, or Panes, or some similar concept — each a little class, instantiated with a Person, and returning the data available on that section view. So changing what appears on the 'Bio' view (for example) should be scoped to only changing that class, and when we want to add a whole new view onto a Person (e.g. info on their assets and expenses) we can add a new class for calculating and returning _that_ data.

tmtmtmtm:
superseded by https://github.com/everypolitician/viewer-sinatra/pull/15119

PR #15119, opened 2016-08-26
WIP Add person card subview classes

ondenman:
Each subview class provides a list of entries for a person card.

screen shot 2016-08-26 at 12 16 10

For example, if `person` is Margaret Beckett, PersonBio.new(person).entries would return:
```
[
    {type: 'Gender', value: 'female},
    {type: 'Born', value: '1943-01-15'},
    {type: 'Prefix', value: 'The Right Honourable'},
]
```

The new classes are: `PersonBio`, `PersonContacts`, `PersonIdentifiers`, `PersonSocial`. They inherit from `PersonCard`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,6 +10,14 @@ Rake::TestTask.new do |t|
 end
 
 Rake::TestTask.new do |t|
+  t.name = 'test:subviews'
+  t.warning = true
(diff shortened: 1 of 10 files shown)


tmtmtmtm:
superseded

PR #15129, opened 2016-08-26
Use webmock instead of vcr

chrismytton:
Using VCR in the tests was leading to hard to track down problems. It was also causing a lot of bloat, since the same popolo files would appear in multiple VCR cassettes. As a solution to these problems this pull request switches to using webmock on its own, with a copy of the fixtures stored locally.


chrismytton:
I'm not totally happy with https://github.com/everypolitician/viewer-sinatra/pull/15129/commits/9b7f0643b60a66280932e2e57433c65c258c3ba3, it seems like there should be a nicer way to test this rather than stubbing 3 calls to Github?

tmtmtmtm:
👍  We'll want to evolve from this, and I fear a few short-term clashes with other open branches, but let's merge, and iterate!

PR #15176, opened 2016-08-27
Add bundle-audit checking

tmtmtmtm:
Use bundler-audit (https://github.com/rubysec/bundler-audit) to report
if any of the gems we're using have known vulnerabilities.

This is integrated into Travis, with auto-updating of the vulnerabilities file.


tmtmtmtm:
@chrismytton do you think this is a good thing to include by default in our projects?

chrismytton:
I hadn't heard of bundler-audit before but it certainly seems like a very sensible thing to have on our other projects. 👍

PR #15245, opened 2016-08-30
Add "House Download" page

octopusinvitro:
## What does this do?

This PR creates dedicated download pages for a single-legislature-data download.

## Why was this needed?

The per-country Download page is too cluttered and confusing, so we want to split it up into separate Download pages per legislature instead.

## Relevant Issue(s)

Part of #13599. Part of #15256.

## Implementation notes

Adds:
* A new page class, `HouseDownload`
* A route in the app.rb
* A new template (copy-pasted the country download one, removed the loop on legislatures and updated the relevant stuff to make it a legislature download page)
* Tests for what we send to the template from the new page

## Screenshots
house

## Notes to Reviewer

There is a lot of duplication here with the Country Download page (particularly the template). However, we will be removing that page before this one goes public, so it's OK to live with the duplication for a little while.

## Notes to Merger

This should not go fully live (i.e. on EveryPolitician.org until all the tickets in the first section of https://github.com/everypolitician/viewer-sinatra/issues/15256 are complete, however it is safe to merge before then as no other public pages will link to here, and so it will not yet get spidered.

### DIFF
--- a/lib/page/house_download.rb
+++ b/lib/page/house_download.rb
@@ -2,12 +2,15 @@
 
 module Page
   class HouseDownload
-    attr_reader :country, :house
+    attr_reader :house
 
-    def initialize(country:, house:, index:)
-      @country = country
-      @house   = house
-      @index   = index
+    def initialize(house:, index:)
+      @house = house
+      @index = index
+    end
+
+    def country
+      house.country
     end
 
     def title
--- a/lib/page/house_download.rb
+++ b/lib/page/house_download.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Page
+  class HouseDownload
+    attr_reader :country, :house
+
+    def initialize(country:, house:, index:)
+      @country = country
+      @house   = house
+      @index   = index
+    end
+
+    def title
+      "EveryPolitician: #{country.name} - #{house.name}"
+    end
+
+    def legislative_periods
+      house.legislative_periods
+    end
+
+    def download_url
+      index.index_url
+    end
+
+    private
+
+    attr_reader :index
+  end
+end
--- a/t/web/house_download.rb
+++ b/t/web/house_download.rb
@@ -45,13 +45,17 @@ describe 'house download template' do
     end
   end
 
-  describe 'term information' do
-    it 'shows the right name' do
+  describe 'terms information' do
+    it 'shows all the terms' do
+      subject.css('li[id*=term-]').count.must_equal 18
+    end
+
+    it 'shows the right name of a term' do
       subject.css('#term-senate-114 h3').text.must_include '114th Congress'
     end
 
-    it 'shows the right dates' do
-      subject.css('#term-senate-114 p').text.must_include '2015-01-06'
+    it 'shows the right dates of a term' do
+      subject.css('#term-senate-113 p').text.strip.must_include '2013-01-06 - 2015-01-03'
     end
   end
 end
--- a/t/web/house_download.rb
+++ b/t/web/house_download.rb
@@ -10,15 +10,15 @@ describe 'house download template' do
     let(:banner)       { subject.css('.page-section--grey h3').text }
     let(:house_header) { subject.css('.country__legislature__header h2').text }
 
-    it 'show the house name in the banner' do
+    it 'shows the house name in the banner' do
       banner.must_include 'Senate'
     end
 
-    it 'show the country name in the banner' do
+    it 'shows the country name in the banner' do
       banner.must_include 'United States of America'
     end
 
-    it 'show the house name in the legislature header' do
+    it 'shows the house name in the legislature header' do
       house_header.must_include 'Senate'
     end
   end
--- a/t/web/house_download.rb
+++ b/t/web/house_download.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../app'
+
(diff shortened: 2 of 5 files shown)


lib/page/house_download.rb
```diff
@@ -0,0 +1,29 @@
...
+      @index   = index
+    end
+
+    def title
+      "EveryPolitician: #{country.name} - #{house.name}"
+    end
+
+    def legislative_periods
+      house.legislative_periods
+    end
```
tmtmtmtm: All the tests pass if I change this implementation to `house.legislative_periods.take(1)`, which probably isn't so good…

t/web/house_download.rb
```diff
@@ -0,0 +1,52 @@
...
+
+describe 'house download template' do
+  subject { Nokogiri::HTML(last_response.body) }
+  before { get '/united-states-of-america/senate/download.html' }
+
+  describe 'headings' do
+    let(:banner)       { subject.css('.page-section--grey h3').text }
+    let(:house_header) { subject.css('.country__legislature__header h2').text }
+
+    it 'show the house name in the banner' do
```
tmtmtmtm: these read a little weirdly — the "it" form exists so that these can read like normal sentences: e.g. "it should show…" or "it shows…" rather than "it show"

t/web/house_download.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    it 'links to the names file' do
+      all_data[1].text.must_include '/United_States_of_America/Senate/names.csv'
+    end
+
+    it 'links to the term file' do
+      last_term.text.must_include '/United_States_of_America/Senate/term-114.csv'
+    end
+  end
```
tmtmtmtm: Isn't the point of this page that it links to _all_ the term files, not *the* term file?

t/web/house_download.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  describe 'term information' do
+    it 'shows the right name' do
+      subject.css('#term-senate-114 h3').text.must_include '114th Congress'
+    end
+
```
tmtmtmtm: ditto here

t/web/house_download.rb
octopusinvitro: Ah yeah, I was reading it as "headings show...", you are right! :+1:

lib/page/house_download.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Page
+  class HouseDownload
+    attr_reader :country, :house
+
+    def initialize(country:, house:, index:)
+      @country = country
+      @house   = house
+      @index   = index
```
tmtmtmtm: Why do you need to instantiate with a `@country` here?

views/house_download.erb
```diff
@@ -0,0 +1,198 @@
+<div class="page-section page-section--grey text-center">
+  <div class="container">
+    <h3>Download data for politicians in <%= @page.house.name %>, <%= @page.country.name %>.</h3>
```
davewhiteland: I think this might be simpler as "Download data for politicians in this legislature." because

* you've already got the house and country name immediately above this line anyway
* it might an article (e.g., "...politicians in *the* Senate..." but I'm not sure that automatically works for every house name we have (it might)

So I suggest you simplify this with no variables in the string, and if it causes upset subsequently raise an issue to fix/refine it.

octopusinvitro: Yeah, as we did in the House page :+1:

octopusinvitro:
house

tmtmtmtm:
Generally looking good, though I'm a little concerned that we have a "Want politicians who aren’t from [this country]?" section, but don't explain what to do if you want politicians from _a different house_ in the same country. It doesn't necessarily need to be solved in this PR, but if it's going to happen separately we should at least link to the issue for resolving that. (Likewise for un-hiding the terms.)

@davewhiteland perhaps this PR is a good test case for rewriting the description in the style that the new PR template should take?

tmtmtmtm:
@davewhiteland I've rewritten the Issue description more in line with the style I think we should move to, and cross-linked the things that are now more clearly able to be done post-merge, than as pre-requisites to merge. So I think this is good to go! 👍 :shipit:

tmtmtmtm:
## Review summary

### Meets acceptance criteria?

* This does not yet meet the criteria for closing #13599 or #15256. It is merely a step towards each of them. But other than some things that are already scheduled to be cleaned up in other tickets, it's doing the right thing.

### UX

Slight issue with getting to equivalent page for other legislatures in this country, but that can be done in another PR.

### Architecture

* one new class — follows existing patterns with no new interface/naming decisions. 
* new Page is instantiated with suitable (and minimal) arguments. 

### Tests
* [x] all new public methods tested
* [x] existing tests all still pass

### Style
* [x] rubocop clean  
    * though adds 7 new "line too long" warnings
* [x] no warnings from flay
* [x] acceptable flog scores 
    * Highest = 10.2 for `HouseDownload#subject` test set-up

### Git/GitHub

* [x] Good PR title
* [x] Good PR description
* [ ] `fixup!`s squashed

@davewhiteland I've rewritten the Issue description more in line with the style I think we should move to, and cross-linked the things that are now more clearly able to be done post-merge, than as pre-requisites to merge. So I think this is good to go! 👍 

@octopusinvitro If you want squash down the fixup! commit before merging, go ahead. Otherwise, let's :shipit:

PR #15246, opened 2016-08-30
House page

octopusinvitro:
This PR adds a new page class for a legislature.

* Adds a `Page::House` class and tests
* Adds the route to app.rb (`/:country/:house/`)
* Adds a new template and tests. The template was taken from the country template which was updated to show information from the house and to remove the loop on legislatures.

house

### DIFF
--- a/lib/page/house.rb
+++ b/lib/page/house.rb
@@ -4,9 +4,9 @@ module Page
   class House
     attr_reader :country, :house
 
-    def initialize(country:, house:)
-      @country = country
+    def initialize(house:)
       @house   = house
+      @country = house.country
     end
 
     def legislative_periods
--- a/lib/page/house.rb
+++ b/lib/page/house.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Page
+  class House
+    attr_reader :country, :house
+
+    def initialize(country:, house:)
+      @country = country
+      @house   = house
+    end
+
+    def legislative_periods
+      house.legislative_periods
+    end
+
+    def title
+      "EveryPolitician: #{country.name} - #{house.name}"
+    end
+  end
+end
--- a/lib/page/house.rb
+++ b/lib/page/house.rb
@@ -2,11 +2,14 @@
 
 module Page
   class House
-    attr_reader :country, :house
+    attr_reader :house
 
     def initialize(house:)
-      @house   = house
-      @country = house.country
+      @house = house
+    end
+
+    def country
+      house.country
     end
 
     def legislative_periods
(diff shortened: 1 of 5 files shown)


lib/page/house.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Page
+  class House
+    attr_reader :country, :house
+
+    def initialize(country:, house:)
+      @country = country
+      @house   = house
+    end
```
tmtmtmtm: you shouldn't need to instantiate with both of these. Passing just the `house` would be cleaner, and the `country` can trivially be derived from that.

t/page/house.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  describe 'legislative periods' do
+    it 'has all legislative periods' do
+      subject.legislative_periods.count.must_equal 18
+    end
+
+    it 'is a list of terms' do
+      subject.legislative_periods.first.slug.must_equal '114'
+    end
```
tmtmtmtm: The test here doesn't seem to match the description. I'm not sure what you're actually trying to test, but "The first things in this list must have a slug of 114" doesn't seem to be a good way to check that you have a list of terms…

views/house.erb
```diff
@@ -0,0 +1,58 @@
+<div class="page-section page-section--grey text-center">
+  <div class="container">
+    <h3>We’re cataloguing every politician in <%= @page.country.name %>.</h3>
+    <p>We currently have information for…</p>
+  </div>
```
tmtmtmtm: This is a little misleading now, because it implies that this is _all_ we have for this country.

views/house.erb
```diff
@@ -0,0 +1,58 @@
...
+</div>
+
+<div class="page-section">
+  <div class="container">
+
+    <div class="country__legislature">
+      <div class="country__legislature__header">
+        <h3><%= @page.house.name %></h3>
```
tmtmtmtm: This seems a little redundant now that there can only be one legislature on the page, and it's already highlighted in the header.

tmtmtmtm:
There's a lot of duplication now between this template and the country one. Unlike https://github.com/everypolitician/viewer-sinatra/pull/15245, where the plan is to delete the per-country download page anyway, resolving the issue, I'm much less convinced it's OK here. We'll have two pages that are almost identical and it's going to be very easy to forget to change one when the other changes.

octopusinvitro:
After rewording the House template, it looks like this:

house

tmtmtmtm:
:+1: — except we can't merge before #15245, as the pages here have a link to the (not yet existing) download page for the legislature.

PR #15247, opened 2016-08-30
Rename route slugs for consistency

octopusinvitro:
We found that there was not a consistent way of naming the slugs in the app routes so we renamed them to be less confusing and consistent.

For example some times there was `cslug`, some times `slug`, some times `country` (the latter was also confusing because there is a `country` that refers to the instance of a `Country` but was also refering to the slug), etc.

All the routes now have their slugs renamed to `country_slug` and `house_slug`.


tmtmtmtm:
Woot! 👍

PR #15251, opened 2016-08-31
Simplify term_table_url

### LINKED ISSUE #15241: term_table_url should only need a Term
Now that we're passing objects everywhere, the `term_table_url` helper shouldn't need to be passed a country or legislature — it can get these from `term.legislature` and `legislature.country`

octopusinvitro:
The `term_table_url` method in the popolo helper was simplify so that it is passed just a term, and the country and legislature are obtained from it using `term.country` and `term.legislature`.

I also changed the bracket notation and used the instance methods (`.slug` and `csv_url`) instead.

Finally, I updated the templates using it, **country.erb**, **country_download.erb** and **term_table.erb**.

Fixes #15241

### DIFF
--- a/lib/popolo_helper.rb
+++ b/lib/popolo_helper.rb
@@ -6,7 +6,7 @@ module Popolo
       '/%s/%s/term-table/%s.html' % [
         term.country.slug.downcase,
         term.legislature.slug.downcase,
-        term.csv_url[/term-(.*?).csv/, 1],
+        term.slug,
       ]
     end
 
--- a/lib/popolo_helper.rb
+++ b/lib/popolo_helper.rb
@@ -2,8 +2,12 @@
 
 module Popolo
   module Helper
-    def term_table_url(c, h, t)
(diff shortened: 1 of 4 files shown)


lib/popolo_helper.rb
```diff
@@ -2,8 +2,12 @@
 
 module Popolo
   module Helper
-    def term_table_url(c, h, t)
-      '/%s/%s/term-table/%s.html' % [c[:slug].downcase, h[:slug].downcase, t[:csv][/term-(.*?).csv/, 1]]
+    def term_table_url(term)
+      '/%s/%s/term-table/%s.html' % [
+        term.country.slug.downcase,
+        term.legislature.slug.downcase,
+        term.csv_url[/term-(.*?).csv/, 1],
```
tmtmtmtm: Is there not a method on `term` that will get us a suitable ID here? If not there really should be…

tmtmtmtm:
Woot! 👍

PR #15252, opened 2016-08-31
WIP -- Reduce complexity of #person_membership

ondenman:
Extracted logic from #person_membership to reduce complexity of method

### DIFF
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -95,12 +95,17 @@ module Page
 
     def person_memberships(person)
       membership_lookup[person.id].each do |mem|
-        group = org_lookup[mem.on_behalf_of_id].first.name if mem.on_behalf_of_id
-        group = '' if group.to_s.downcase == 'unknown'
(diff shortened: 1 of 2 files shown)


lib/page/term_table.rb
```diff
@@ -95,14 +95,19 @@ def top_identifiers
...
-        group = org_lookup[mem.on_behalf_of_id].first.name if mem.on_behalf_of_id
-        group = '' if group.to_s.downcase == 'unknown'
         area = area_lookup[mem.area_id].first.name if mem.area_id
+        group = member_group(mem)
         mem.define_singleton_method(:group) { group || '' }
         mem.define_singleton_method(:area)  { area  || '' }
       end
     end
 
+    def member_group(member)
```
tmtmtmtm: This name is at best confusing, and at worst actively misleading. I would expect this to be given a Member, and return a Group. But neither half of that is true…

(I strongly suspect that finding the right name for it will reveal that it should live somewhere else entirely…)

ondenman:
Closing as restarted on new branch.

PR #15253, opened 2016-08-31
WIP -- Reduce complexity of percentages method

ondenman:
PR reduces complexity of #percentages by assigning value of `people.count` to variable so that `people.count` is called once only. Method now passes Rubocop's ABC test.


ondenman:
Closing. This doesn't solve the problem.

PR #15255, opened 2016-08-31
Link to house/download

octopusinvitro:
In the Country view, make the download links in every legislature section point to their legislature-download dedicated page.

This PR depends on #15245 being merged


tmtmtmtm:
👍

PR #15257, opened 2016-08-31
WIP -- Create stats class

ondenman:
Moving logic from term-table.rb to separate class.

### DIFF
--- a/lib/term_statistics.rb
+++ b/lib/term_statistics.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+class TermStatistics
+  def initialize(term:, org_lookup:)
+    @term = term
+    @org_lookup = org_lookup
+  end
+
+  SeatCount = Struct.new(:group_id, :name, :member_count)
+  def group_data
+    @group_data ||= term
+                    .memberships_at_end
+                    .group_by(&:on_behalf_of_id)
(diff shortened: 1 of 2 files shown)


lib/term_statistics.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+class TermStatistics
+  # TODO: reduce number of arguments
+  def initialize(term:, org_lookup:, people:)
```
tmtmtmtm: I'm not sure what we want to do generally about documentation, but I think we definitely need to do _something_ to explain what these arguments are. (And I strongly suspect that doing so would reveal `org_lookup`, at least, isn't really going to be a suitable thing to be passing around)

tmtmtmtm: Hmmm. After looking a little closer, I don't think `people` is obvious here either. I would naturally have assumed that that was going to be a list of `EveryPolitician::Person` objects, but nope!

tmtmtmtm:
I'm afraid I'm finding it really hard to puzzle out what this class represents. In particular, I can't easily tell what it gives me (and what I need to give it to get that). I suspect that 90% of this is a question of documentation, but I also suspect that writing that documentation will prove difficult enough to show that they might be wrong. (Adding unit tests for this may also show likewise…)

PR #15258, opened 2016-08-31
Test group_data for empty array

ondenman:
#group_data should return an empty array when there is only a single group for a given term. Added test to test that.


tmtmtmtm:
👍

PR #15260, opened 2016-08-31
WIP Extend Popolo::Memberships

ondenman:
Adding #group, #group_name, #area, #area_name to class

### DIFF
--- a/lib/everypolitician_extensions.rb
+++ b/lib/everypolitician_extensions.rb
@@ -38,8 +38,27 @@ module EveryPolitician
       end
     end
   end
+
+  module MembershipExtension
+    def group
+      popolo.organizations.find_by(id: on_behalf_of_id)
+    end
+
+    def area
+      popolo.areas.find_by(id: area_id)
+    end
+
+    def group_name
+      group.name.to_s.downcase == 'unknown' ? '' : group.name
+    end
+
+    def area_name
+      area.name if area_id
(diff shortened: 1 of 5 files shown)


lib/everypolitician_extensions.rb
```diff
@@ -38,8 +38,27 @@ def memberships_at_end
...
+      popolo.organizations.find_by(id: on_behalf_of_id)
+    end
+
+    def area
+      popolo.areas.find_by(id: area_id)
+    end
+
+    def group_name
+      group.name.to_s.downcase == 'unknown' ? '' : group.name
+    end
```
tmtmtmtm: This one feels wrong; or at least in the wrong place. What's the thinking here?

lib/everypolitician_extensions.rb
```diff
@@ -38,8 +38,27 @@ def memberships_at_end
...
+
+    def area_name
+      area.name if area_id
+    end
+  end
 end
```
tmtmtmtm: I'm a little hesitant about this one, too. Are we going to create equivalent methods like this for every other method that someone might call on an Area? If we really need something like this, then I suspect we want `def area` to return some sort of NullArea if none exists.

ondenman:
Closed. Superseded by https://github.com/everypolitician/viewer-sinatra/pull/15323

PR #15261, opened 2016-08-31
Country download redirect

octopusinvitro:
This PR is superseded by #15464 and #15466
----
This PR redirects the country/download.html route to the country page.

* The `redirect.erb` template was used for both the redirection to the docs and country pages. Since the url of the host was hardcoded in the view, it is now passed as a variable to it [1].

* New tests were added to ensure that the redirection to both the docs pages and the country page is working properly (`/t/web/redirect.rb`).

* Since the docs and the country page have different hosts, I updated the signature of the redirect method to take the host as an argument. The `/about.html` route was updated as well

* To mirror `settings.doc_url` I added `set :main_url, 'http://everypolitician.org'` to be used by the country/download redirection

* The country/download page class, view and page/web tests were deleted.

---------------------------
[1] Passing `@url` to the view also allowed to get rid of the non-existing `@filename` variable referenced in said template. The tests were showing that when hitting a documents route (for example, `everypolitician.org/contribute.html`), the presence of this non-existing variable was causing a redirection to `docs.everypolitician.org` instead of `docs.everypolitician.org/contribute.html`.


tmtmtmtm:
@octopusinvitro this PR is a little too large, especially as it has a dependency on other PRs that haven't been merged yet. I'd suggest splitting out the parts that refactor and test the existing code (and which make _this_ change easy), leaving this PR as the part that can be the subsequent easy change.

app.rb
```diff
@@ -118,11 +119,11 @@
 # because wget simply fetches the HTML document, this lets us continue to
 # spider the site to generate the contents of everypolitician/viewer-static.
 # See scripts/release.sh (update_viewer_static).
-def docs_redirect(path, page_title)
-  url = URI.join(settings.docs_url, path)
+def redirect(host, path, page_title)
```
tmtmtmtm: It seems a little odd to pass a host and path separately when you're only ever going to use them together. Perhaps better to simply require a URI?

app.rb
```diff
@@ -88,7 +89,7 @@
 
 # Old doc pages are now at docs.everypolitician.org: redirect to them
 get '/about.html' do
-  docs_redirect('/', 'About')
+  redirect(settings.docs_url, '/', 'About')
```
tmtmtmtm: I think I'd leave a separate `docs_redirect` method that in turn calls the more general redirection method with a URI constructed with `settings.docs_url`, rather than having to change the existing callers.

app.rb
tmtmtmtm: Calling this `redirect` scares me a little, as it's doing something different from what most people are likely to think a `redirect` method would do. I think it would be better to make it more explicitly `soft_redirect` or `meta_redirect` or somesuch.

PR #15262, opened 2016-08-31
First steps towards documentation

tmtmtmtm:
Add YARD, and add some initial documentation of a single class (Page::Home). 

There are undoubtedly lots of ways to improve on this, but this is initially just seeing how the basic approach works before rolling it out everywhere.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .env
 .vagrant
 .sass-cache
+.yardoc/
 
 *.swo
 
--- a/Gemfile
+++ b/Gemfile
@@ -29,4 +29,5 @@ group :quality do
   gem 'flog'
   gem 'reek'
   gem 'rubocop'
+  gem 'yard'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -111,6 +111,7 @@ GEM
       crack (>= 0.3.2)
       hashdiff
     yajl-ruby (1.2.1)
+    yard (0.9.5)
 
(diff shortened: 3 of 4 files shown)


tmtmtmtm:
Running `yard server --reload` is useful for generating the documentation whilst creating/editing it, but it's not totally clear to me how we should actually publish this normally. Are we meant to run `yard doc` and then check that in?

tmtmtmtm:
@chrismytton thoughts? suggestions? ideas? Should we be using something else instead?

chrismytton:
> Running yard server --reload is useful for generating the documentation whilst creating/editing it, but it's not totally clear to me how we should actually publish this normally. Are we meant to run yard doc and then check that in?

You can see the generated documentation on [rubydoc.info](http://www.rubydoc.info/github/everypolitician/viewer-sinatra/yard/Page/Home) (that link is for this `yard` branch)

> @chrismytton thoughts? suggestions? ideas? Should we be using something else instead?

Big 👍  from me! We might also want to add [`yard-sinatra`](https://github.com/rkh/yard-sinatra) (semi-official, since it's written by Sinatra's primary maintainer) which adds Sinatra routes to the documentation. The main alternatives are TomDoc and RDoc, but neither of those have as much momentum in the Ruby community AFAIK.

tmtmtmtm:
Good call on yard-sinatra: ticket for that opened at https://github.com/everypolitician/viewer-sinatra/issues/15306

PR #15323, opened 2016-09-01
Extend Popolo::Membership with #group and #area

ondenman:
Methods return group and area for a given member.

TermTable currently uses `#org_lookup` and `#area_lookup` to get the area and group for a given member. The intention is for the class to use the extension methods instead.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,6 +20,15 @@ Rake::TestTask.new do |t|
 end
 
 Rake::TestTask.new do |t|
+  t.name = 'test:extensions'
+  t.warning = false
+  t.verbose = true
+  t.description = 'Run "Everypolitician extensions" tests'
+  t.test_files = FileList['t/everypolitician_extensions/**/*.rb']
+  t.libs << 't'
+end
+
+Rake::TestTask.new do |t|
   t.name = 'test:all'
   t.verbose = true
   t.description = 'Run all tests (slow)'
--- a/lib/everypolitician_extensions.rb
+++ b/lib/everypolitician_extensions.rb
@@ -38,8 +38,19 @@ module EveryPolitician
       end
     end
   end
+
+  module MembershipExtension
+    def group
+      popolo.organizations.find_by(id: on_behalf_of_id)
+    end
+
+    def area
+      popolo.areas.find_by(id: area_id)
+    end
+  end
 end
 
 EveryPolitician::Country.include EveryPolitician::CountryExtension
 EveryPolitician::Legislature.include EveryPolitician::LegislatureExtension
 EveryPolitician::LegislativePeriod.include EveryPolitician::LegislativePeriodExtension
+EveryPolitician::Popolo::Membership.include Everypolitician::MembershipExtension
--- a/lib/everypolitician_extensions.rb
+++ b/lib/everypolitician_extensions.rb
@@ -40,10 +40,14 @@ module EveryPolitician
   end
 
   module MembershipExtension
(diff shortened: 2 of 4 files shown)


lib/everypolitician_extensions.rb
```diff
@@ -38,8 +38,23 @@ def memberships_at_end
       end
     end
   end
+
+  module MembershipExtension
+    # The group that a membership represents
+    # @return [<Everypolitician::Popolo::Organization>]
```
tmtmtmtm: I don't think you want the angle brackets there (or on `area`)

lib/everypolitician_extensions.rb
```diff
@@ -38,8 +38,23 @@ def memberships_at_end
       end
     end
   end
+
+  module MembershipExtension
+    # The group that a membership represents
```
tmtmtmtm: This is a slightly odd description. A membership doesn't really _represent_ a Group (or Area in version below). 

I'd suggest something like:
* The political group (party / faction) this membership is held on behalf of.

Rakefile
```diff
@@ -20,6 +20,15 @@ Rake::TestTask.new do |t|
 end
 
 Rake::TestTask.new do |t|
+  t.name = 'test:extensions'
+  t.warning = false
+  t.verbose = true
+  t.description = 'Run "Everypolitician extensions" tests (slow)'
```
tmtmtmtm: Why are these marked as '(slow)'?

t/everypolitician_extensions/membership_extension.rb
```diff
@@ -0,0 +1,32 @@
...
+describe 'MembershipExtensions' do
+  subject do
+    stub_popolo('3df153b', 'Austria/Nationalrat')
+    index_at_known_sha.country('austria')
+                      .legislature('nationalrat')
+                      .term('25')
+                      .legislature
+                      .popolo
+                      .memberships
+                      .first
```
tmtmtmtm: I don't think you need so much indirection from legislature to term and back to legislature again there. `country('austria').legislature('nationalrat').popolo.memberships.first` should be enough.

t/everypolitician_extensions/membership_extension.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  it 'should return nil if a membership has no group data' do
+    stub_popolo('2b38667', 'Afghanistan/Wolesi_Jirga')
+    subject = index_at_known_sha.country('afghanistan')
+                                .legislature('wolesi-jirga')
+                                .popolo
+                                .memberships
+                                .first
```
tmtmtmtm: It's not clear why you're providing a different `subject` here to the one that you've pre-defined above. If you really need one, then it would be better to put it up to a `let` definition. But I suspect this can just be deleted, as it looks identical.

ondenman: Removed in fixup.

tmtmtmtm:
👍  

(squashed fixups before merging)

t/everypolitician_extensions/membership_extension.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  it 'should return nil if a membership has no group data' do
+    subject.area.must_be_nil
+  end
+
+  it 'should return "unkown" if membership’s area name is unknown' do
+    subject.group.name.must_equal 'unknown'
+  end
+end
```
tmtmtmtm: Bah. Only noticed this after merging, but these descriptions seem to be the wrong way around…

PR #15325, opened 2016-09-01
 Update to the new signature of term_table_url

octopusinvitro:
After rebasing master onto this branch and running the tests, some were failing due to the recent simplification of the signature of the term_table_url method. This commit updates the template to use the new signature.


tmtmtmtm:
👍

PR #15428, opened 2016-09-02
Simplify percentages method

ondenman:
Code in lambda block that contains percentage of members with biographical/social/contact/identifiers data moved to separate method.


ondenman:
Closing. Tackling in new branch.

PR #15431, opened 2016-09-02
Fix behaviour descriptions

ondenman:
Descriptions were named wrong. One was testing the area but described as testing the group and vice versa.


tmtmtmtm:
👍

PR #15432, opened 2016-09-02
Data completeness subview class

ondenman:
Added class to provide data for data completeness subview:

screen shot 2016-09-02 at 12 30 12

Extracted from `TermTable#percentages`

### DIFF
--- a/lib/data_completeness.rb
+++ b/lib/data_completeness.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+class DataCompleteness
+  def initialize(people:)
+    @people = people
+  end
+
+  CARDS = %i(social bio contacts identifiers).freeze
+  Percentages = Struct.new(*CARDS)
+  def percentages
+    Percentages.new(*CARDS.map { |card| calculate_completeness(card) })
+  end
+
+  private
+
+  attr_reader :people
+
+  def calculate_completeness(card)
+    ((people.count { |p| p.send(card.to_s).any? } / people.count.to_f) * 100).floor
+  end
+end
--- a/lib/data_completeness.rb
+++ b/lib/data_completeness.rb
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
+# The completeness of each type of member data as a percentage
 class DataCompleteness
+  # @param [Array PersonCard]
   def initialize(people:)
     @people = people
   end
 
   CARDS = %i(social bio contacts identifiers).freeze
   Percentages = Struct.new(*CARDS)
+  # The percentage of data completeness for categories defined in CARDS
+  # Return [<struct DataCompleteness::Percentages>]
   def percentages
     Percentages.new(*CARDS.map { |card| calculate_completeness(card) })
   end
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
Yardoc is complaining about:

```
[warn]: @param tag has unknown parameter name:  
    in file `lib/data_completeness.rb' near line 5
```

lib/data_completeness.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+# The completeness of each type of member data as a percentage
+class DataCompleteness
+  # @param [Array PersonCard]
+  def initialize(people:)
```
tmtmtmtm: If this is receiving an array of` `PersonCard`, perhaps better to call this something like `person_cards`. `people` would imply to me that it's wanting `Person` objects instead.

lib/data_completeness.rb
```diff
@@ -0,0 +1,24 @@
...
+    Percentages.new(*CARDS.map { |card| calculate_completeness(card) })
+  end
+
+  private
+
+  attr_reader :people
+
+  def calculate_completeness(card)
+    ((people.count { |p| p.send(card.to_s).any? } / people.count.to_f) * 100).floor
+  end
```
tmtmtmtm: This is really a query method, but it's named as if it's a command method. (See, for example, http://martinfowler.com/bliki/CommandQuerySeparation.html)

Methods that are returning you something should simply be named for what they return, not what they _do_ to return that — so here just calling this `completeness` would be better.

lib/page/term_table.rb
```diff
@@ -2,6 +2,7 @@
 require 'everypolitician'
 require_relative '../everypolitician_extensions'
 require_relative '../person_cards'
+require_relative '../data_completeness.rb'
```
tmtmtmtm: Better to keep consistent with the others here, and not include the `.rb`

tmtmtmtm:
@ondenman You haven't fixed the problem yet that yardoc is complaining about at line 5 (i.e. you haven't documented what the variable is called — only its type)

lib/data_completeness.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+# The completeness of each type of member data as a percentage
+class DataCompleteness
+  # @param [Array<PersonCard>]
+  def initialize(person_cards:)
+    @person_cards = person_cards
+  end
+
+  CARDS = %i(social bio contacts identifiers).freeze
+  Percentages = Struct.new(*CARDS)
```
tmtmtmtm: This `Percentages` class isn't documented…

lib/data_completeness.rb
```diff
@@ -0,0 +1,24 @@
...
+  # The percentage of data completeness for categories defined in CARDS
+  # @return [struct DataCompleteness::Percentages]
```
tmtmtmtm: Did you see somewhere that you should be including this `struct` here? I can't see anything in http://www.rubydoc.info/gems/yard/file/docs/Tags.md about this, though I could easily be overlooking something

PR #15433, opened 2016-09-02
Add html-validation to detect malformed markup

octopusinvitro:
Firefox was showing the source code with the heading tags in red. I discovered that it was due to a missing link-closing tag, so I added </a> to the main layout heading. But it is much better if we have an automated way to detect and correct those things. Hence this addition. The gem uses the [tidy-html5](https://github.com/htacg/tidy-html5) validator, which can be installed locally optionally.

I ran the validator against each page and did a separate commit for each, including the reported errors for that page.

@zarino , I had to remove some of your code (basically attributes in the main page's `select`) and add code on some other places (mostly `alt` attributes to images), so I would like you to take a look at this as well if you please.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ group :test do
   gem 'rack-test'
   gem 'webmock'
   gem 'bundler-audit'
+  gem 'html_validation'
 end
 
 group :quality do
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -305,6 +305,11 @@ $(function(){
   });
 
   // http://baymard.com/labs/country-selector
+  $('.js-select-to-autocomplete').attr({
+    placeholder:  'Search by country name',
+    autocorrect:  'off',
+    autocomplete: 'off'
+  });
   $('.js-select-to-autocomplete').selectToAutocomplete().on('change', function(){
     var v = $(this).val();
     if (v) {
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -305,20 +305,23 @@ $(function(){
   });
 
   // http://baymard.com/labs/country-selector
-  $('.js-select-to-autocomplete').attr({
-    placeholder:  'Search by country name',
-    autocorrect:  'off',
-    autocomplete: 'off'
-  });
-  $('.js-select-to-autocomplete').selectToAutocomplete().on('change', function(){
-    var v = $(this).val();
-    if (v) {
-      // Assumes the <option>'s `value` attribute is a URL slug for the country
-      window.location.href = '/' + v + '/';
-    }
-  }).on('focus', function(){
-    $(this).next().trigger("focus");
-  });
+  $('.js-select-to-autocomplete')
+    .attr({
+      placeholder:  'Search by country name',
+      autocorrect:  'off',
+      autocomplete: 'off'
+    })
+    .selectToAutocomplete()
+    .on('change', function(){
+      var v = $(this).val();
+      if (v) {
+        // Assumes the <option>'s `value` attribute is a URL slug for the country
+        window.location.href = '/' + v + '/';
+      }
+    })
+    .on('focus', function(){
+      $(this).next().trigger("focus");
+    });
 
   // Fix the incorrect default autocomplete width, which meant that
   // autocomplete menu was longer than the search input it's linked to.
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -140,10 +140,13 @@
                             <img src="/images/person-placeholder-108px.png"
                                  data-src="<%= person.proxy_image %>"
                                  style="display: none"
-                                 class="person-card__image">
-                            <noscript><img src="<%= person.proxy_image %>" class="person-card__image"></noscript>
+                                 class="person-card__image"
+                                 alt="Member headshot">
+                            <noscript><img src="<%= person.proxy_image %>" class="person-card__image"
+                            alt="Member headshot"></noscript>
                           <% else %>
-                            <img src="/images/person-placeholder-108px.png" class="person-card__image">
+                            <img src="/images/person-placeholder-108px.png" class="person-card__image"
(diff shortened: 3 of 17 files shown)


tmtmtmtm:
A bugfix without a regression test… surely not? ;)

octopusinvitro:
@tony should I research into adding something that runs a check against the W3C validator when running the web tests

tmtmtmtm:
I don't think you need to run against an external validator. Nokogiri in strict mode is probably all we need.

Changing the `subject` to `Nokogiri::HTML(last_response.body) { |config| config.strict }` lets us then run a test later like

```ruby
it "has no HTML errors" do
   subject.errors.must_be_empty
end
```

zarino:
@octopusinvitro – This looks fine to me, **apart from** the “Search by country name” hint text now being missing from the search input on the homepage.

Before:

screen shot 2016-09-20 at 17 24 35

After:

screen shot 2016-09-20 at 17 24 45

The text input (generated by [selectToAutocomplete](http://baymard.com/labs/country-selector)) uses `copy-attributes-to-text-field` to soak up whatever HTML attributes the source `<select>` element had. I guess, a satisfactory solution if we still want W3C compliant markup, would be to add the `placeholder` attribute in with JavaScript, immediately before we call `. selectToAutocomplete()` in [main.js](https://github.com/everypolitician/viewer-sinatra/blob/75f3218b87e41d26e97b6ad17550630f2af5bdae/public/javascript/main.js#L308).

Gemfile
```diff
@@ -31,5 +31,6 @@ group :quality do
   gem 'flog'
   gem 'reek'
   gem 'rubocop'
+  gem 'ruby-vnu'
```
tmtmtmtm: As this is required to be able to run `rake test:all`, I think it should be in the `:test` group

octopusinvitro:
@zarino I'll do that :+1: 

@tmtmtmtm Tony I totally agree with that, but wouldn't we also forget to add an HTML validation test if we create a new test file from scratch for a new page. I think it is a good call that we have to find a way to ensuring that we remember to add an HTML validation test, I'll think about that. 

Something that is also important but we haven't thought about, I think it would also be a good idea to ensure accessibility, but that's for another day probably.

tmtmtmtm:
> I totally agree with that, but wouldn't we also forget to add an HTML validation test if we create a new test file from scratch for a new page

Possibly, though in general it's much easier for a reviewer to notice if something is missing from a file that's actually _in_ a PR, than to realise that an entire file is missing. (Plus we could fairly easily add some static analysis to ensure that every test file has a suitable test like this, even if just by insisting on naming convention (though we could probably do something cleverer))

octopusinvitro:
@tmtmtmtm I added Chris commits which switch from `ruby-vnu` to `html-validation` and make Travis ~~great~~ green again.

Since I think that is already enough overhead for a single PR I think we shouldn't add more changes to this PR and instead do the fix proposed by Zarino and the separation of the files in different PRs. 

I have already added two issues to cover that, which I can tackle as soon as this PR is merged:
* Make the hint text visible again https://github.com/everypolitician/viewer-sinatra/issues/15562
* Move HTML validation tests together with the pages tests https://github.com/everypolitician/viewer-sinatra/issues/15563

octopusinvitro:
@tmtmtmtm I rebased all commits in the PR to:
* Remove the commits that were adding the old nu-validator
* Remove the commit that validates the home page, and leave that to another PR where  #15562 is also addressed
* Moved the tests to their dedicated pages and remove the html_validation test file
* I had to add a `t/web/countries.rb` page because it didn't exist.

Closes #15563

public/javascript/main.js
```diff
@@ -305,6 +305,11 @@ $(function(){
   });
 
   // http://baymard.com/labs/country-selector
+  $('.js-select-to-autocomplete').attr({
```
zarino: A small thing, but it’s generally good practice to avoid querying the DOM more often than is necessary. So, you might query the DOM once and then cache the result in a variable to be reused again later:

    var $select = $('.js-select-to-autocomplete');
    $select.attr({ … });
    $select.selectToAutoComplete();
    $select.on('change', function(){ … });

Or, since you have a fairly small number of methods to call, you could save the overhead of another variable, and just use jQuery chaining:

    $('.js-select-to-autocomplete')
      .attr({ … })
      .selectToAutocomplete()
      .on('change', function(){ … })

octopusinvitro:
@tmtmtmtm feedback addressed

views/term_table.erb
```diff
@@ -140,10 +140,13 @@
                             <img src="/images/person-placeholder-108px.png"
                                  data-src="<%= person.proxy_image %>"
                                  style="display: none"
-                                 class="person-card__image">
-                            <noscript><img src="<%= person.proxy_image %>" class="person-card__image"></noscript>
+                                 class="person-card__image"
+                                 alt="Member headshot">
```
tmtmtmtm: Is a plain "Member headshot" sufficient, when there might be hundreds of these across the page? Perhaps we should include the member's name here?

views/term_table.erb
```diff
@@ -140,10 +140,13 @@
...
+                            <noscript><img src="<%= person.proxy_image %>" class="person-card__image"
+                            alt="Member headshot"></noscript>
                           <% else %>
-                            <img src="/images/person-placeholder-108px.png" class="person-card__image">
+                            <img src="/images/person-placeholder-108px.png" class="person-card__image"
+                            alt="Placeholder image">
```
tmtmtmtm: ditto here. "Placeholder image for Freddy McFreedy" seems like it might be more useful.

octopusinvitro:
Closing this as it was handled in separate PRs for every page and the only ones pending are https://github.com/everypolitician/viewer-sinatra/pull/15577 and https://github.com/everypolitician/viewer-sinatra/pull/15580

Issue #15256, opened 2016-08-31
release house and house/download pages


davewhiteland:
# Problem

We currently have a Page for a Country, and a parallel Download page for that Country. These group together information for each Legislature in that Country. This gets confusing at times (particularly on the Download page), because we're cramming quite a lot onto them. Instead, we should create similar Pages at a per-Legislature level.

We will need to keep the primary per-Country page (as it's the primary entry point for a Country from various places), but we should be able to completely delete the per-Country download page (with suitable redirects in place).

We should, as normal, deliver this in small steps, but there are various mutual dependencies (as each of these pages will need to link to each other).

This issue captures each of the things that we know we need to do:
# Proposed Solution

Suggested that we release in this order:
### Step 1: New Pages (not yet linked)
- [x] Create a house/download page (#15245)
- [x] house page is published (#15246)
- [x] Ensure house/download links to House page (#15339)
- [x] Ensure House page links to Download page
- [ ] improve text about "other politicians in same country" at bottom of house page ~~#15259~~ (#15451)
### Step 2: Usable Live Pages (would like this week)
- [x] link to Legislature page from Country page #15255 
- [x] link to Legislature page from Term Table page (header bar)
- [ ] stop collapsing terms on download page #15243 
### Step 3: Post-live cleanup (should happen soon, but OK for there to be some delay)
- [ ] check no links exist/remain to country/download pages
- [ ] delete country/download pages and redirect (client-side: because gh-pages)) ~~#15261~~ (#15464 and #15466)
# Acceptance Criteria
- [ ] Working house and house/download pages, linked from Country and Term pages.
- [ ] No broken links on viewer-sinatra site and therefore on the static site.
# Not Required

n/a
# Prerequisites

See issues above.
# Related Issues

See issues above.
# Due Date

n/a (ASAP)
# Design?

No
# Text?

No
# Bloggable?

No

PR #15464, opened 2016-09-03
Test redirect

octopusinvitro:
This PR supersedes [Country download redirect #15261](#15261)

## What does this do?

This PR is the first part of splitting [Country download redirect #15261](#15261) in two. Second part here: #15466

It adds tests for the existing redirecting code (and which make the change at hand easy), leaving the second part of that PR as the part that can be the subsequent easy change.

## Why was this needed?

[Country download redirect #15261](#15261) was a little too large, especially as it had a dependency on other PRs that weren't merged at the moment of its creation.

## Relevant Issue(s)

This is part of issue [release house and house/download pages #15256](#15256) regarding the post-live clean-up, and TO-DO list [TODO w/c 2016-08-30 #492](https://github.com/everypolitician/everypolitician/issues/492)

## Implementation notes

* Since the url of the host was hardcoded in the view, it is now passed as a variable to it [1].

* New tests were added to ensure that the redirection to the docs pages is working properly (`/t/web/redirect.rb`).

---------------------------
[1] Passing `@url` to the view also allowed to get rid of the non-existing `@filename` variable referenced in said template. The tests were showing that when hitting a documents route (for example, `everypolitician.org/contribute.html`), the presence of this non-existing variable was causing a redirection to `docs.everypolitician.org` instead of `docs.everypolitician.org/contribute.html`.

## Screenshots

Not needed

## Notes to Reviewer

Nope

## Notes to Merger

This PR should be merged before #15466

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -133,10 +133,10 @@ end
 # spider the site to generate the contents of everypolitician/viewer-static.
 # See scripts/release.sh (update_viewer_static).
 def docs_redirect(path, page_title)
-  url = URI.join(settings.docs_url, path)
+  @url = URI.join(settings.docs_url, path)
   @head_tags = [
-    %(<meta http-equiv="refresh" content="0; url=#{url}">),
-    %(<link rel="canonical" href="#{url}"/>),
+    %(<meta http-equiv="refresh" content="0; url=#{@url}">),
+    %(<link rel="canonical" href="#{@url}"/>),
   ].join("\n\t")
   @page_title = page_title
   erb :redirect
--- a/t/web/redirect.rb
+++ b/t/web/redirect.rb
@@ -3,12 +3,13 @@ require 'test_helper'
 require_relative '../../app'
 
 describe 'when hitting a docs page' do
-  subject { Nokogiri::HTML(last_response.body) }
-  before { get '/contribute.html' }
+  let(:url) { 'http://docs.everypolitician.org/contribute.html' }
+  subject   { Nokogiri::HTML(last_response.body) }
+  before    { get '/contribute.html' }
 
   it 'should redirect to the right url' do
-    subject.css('meta/@content').first.text.must_equal '0; url=http://docs.everypolitician.org/contribute.html'
-    subject.css('link/@href').first.text.must_equal 'http://docs.everypolitician.org/contribute.html'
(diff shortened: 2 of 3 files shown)


tmtmtmtm:
@mael I don't understand your comment about http://everypolitician.org/contribute.html redirecting to the wrong place. When I go to that it (correctly) goes to http://docs.everypolitician.org/contribute.html

octopusinvitro:
It was not doing so in localhost... nor in the tests.

Where does `@filename` come from? It is not sent to the template explicitly from the `docs_redirect`:
```ruby
def docs_redirect(path, page_title)
  url = URI.join(settings.docs_url, path)
  @head_tags = [
    %(<meta http-equiv="refresh" content="0; url=#{url}">),
    %(<link rel="canonical" href="#{url}"/>),
  ].join("\n\t")
  @page_title = page_title
  erb :redirect
end
```

Example call:
```ruby
get '/about.html' do
  docs_redirect('/', 'About')
end
```

And the template:
```erb
<p class="lead">Redirecting to <a href="http://docs.everypolitician.org/<%= @filename %>">docs.everypolitician.org</a></p>
<p class="fourohfour-suggestions">
  <a href="/" class="button button--primary">Go home</a>
  <span class="fourohfour-suggestions__or">or</span>
  <a href="http://docs.everypolitician.org/<%= @filename %>" class="button button--secondary">Load page</a>
</p>
```

So the behaviour I was experimenting was the one I was expecting.

tmtmtmtm:
@octopusinvitro Ah, I think we're talking about different things. The text that you're showing there is pretty much irrelevant to the functionality. That's something no-one should ever see, unless their browser doesn't automatically follow the meta refresh tag. 

So, yes, it would be good to have it show the right thing, but the part that's relevant to the actual redirection is the 
```
<meta http-equiv="refresh" content="0; url=http://docs.everypolitician.org/contribute.html"
```

line, not the 

```
<p class="lead">
```

one.

octopusinvitro:
You're right...
That template is a bit confusing!
:+1:

tmtmtmtm:
Looks like `@filename` was [never](https://github.com/everypolitician/viewer-sinatra/commit/5c401a41cc30080f209bf909302cf142748be9f4) passed to the template (cc: @davewhiteland)

PR #15547, opened 2016-09-15
Read DATASOURCE from an environment variable rather than a file

chrismytton:
This changes viewer-sinatra to read the `DATASOURCE`, which is a url pointing to the `countries.json` file to use to get the data for the site, from an environment variable instead of a file.

As part of [changing the way everypolitician.org is deployed](https://github.com/everypolitician/everypolitician/projects/2) we want to be able to have a site builder gem that when invoked uses viewer-sinatra to build everypolitician.org using a specific version of `countries.json`. This pull request means that we can specify which version of `countries.json` we want to use without having to make a commit to viewer-sinatra.

Fixes https://github.com/everypolitician/everypolitician/issues/503

### DIFF
--- a/t/page/download.rb
+++ b/t/page/download.rb
@@ -21,10 +21,6 @@ describe 'Download' do
       it 'should be at the correct SHA' do
         subject.download_url.must_include 'd8a4682f'
       end
-
-      it 'should be at rawgit' do
-        subject.download_url.must_equal index_at_known_sha.index_url
-      end
     end
   end
 end
--- a/t/page/download.rb
+++ b/t/page/download.rb
@@ -23,7 +23,7 @@ describe 'Download' do
       end
 
       it 'should be at rawgit' do
-        subject.download_url.must_include 'cdn.rawgit.com'
+        subject.download_url.must_equal index_at_known_sha.index_url
       end
     end
   end
--- a/t/page/house_download.rb
+++ b/t/page/house_download.rb
@@ -44,9 +44,5 @@ describe 'HouseDownload' do
     it 'should be at the correct SHA' do
       subject.download_url.must_include 'd8a4682f'
     end
-
-    it 'should be at rawgit' do
-      subject.download_url.must_equal index_at_known_sha.index_url
-    end
   end
 end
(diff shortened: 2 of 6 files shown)


t/page/download.rb
```diff
@@ -23,7 +23,7 @@
       end
 
       it 'should be at rawgit' do
-        subject.download_url.must_include 'cdn.rawgit.com'
+        subject.download_url.must_equal index_at_known_sha.index_url
```
tmtmtmtm: If this is now longer actually at rawgit, this is now a rather confusingly named test, that I suspect someone will stumble over at some point. At the very least we should rename this test, but I'm wondering if it's even needed at all. If this test started failing, what would it mean?

t/page/house_download.rb
```diff
@@ -46,7 +46,7 @@
     end
 
     it 'should be at rawgit' do
-      subject.download_url.must_include 'cdn.rawgit.com'
+      subject.download_url.must_equal index_at_known_sha.index_url
```
tmtmtmtm: ditto for this one.

t/test_helper.rb
```diff
@@ -58,15 +58,15 @@ def stub_github_api
     private
 
     def cdn
-      'https://cdn.rawgit.com'
+      'https://github.com'
```
tmtmtmtm: It feels a little odd for this to be `cdn` now. This is definitely the easy change to make everything Just Work™, but the mismatch here seems like it might be a little confusing later. It's not a show-stopper, but it certainly causes me some hesitation. Is this used directly from tests, or just locally here? If the latter, I'd be very tempted to just eliminate this method now.

chrismytton: Looks like it's just in `test_helper.rb` so I'll zap it ⚡

t/page/download.rb
chrismytton: I think you're right about removing this, it's essentially an alias for `index.index_url`, so I don't think there's any benefit to be had in keeping this test. I'll get rid of it.

tmtmtmtm:
@davewhiteland @ondenman @octopusinvitro FYI. When you want to run a local version of viewer-sinatra now to check how your data looks, you'll need to pass the countries.json link in the environment, rather than updating the DATASOURCE file.

PR #15548, opened 2016-09-15
add a pull request template

davewhiteland:
General topics -- will solicit refinement in PR

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,16 @@
+# What does this do?
+
+# Why was this needed?
+
+# Relevant Issue(s)
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+
+# Why was this needed?
+
+# Relevant Issue(s)
+
+# Implementation notes
+
+# Screenshots
+
+# Not needed
```
tmtmtmtm: This one is a little too unclear, I think. I'm not sure what I would expected to write here. Worded like this, it feels more like something that's part of describing the Issue, rather than the PR. What would you expect to be covered here that isn't in the first two sections, or the "notes to reviewer"?

PR #15555, opened 2016-09-20
Link to everypolitician-data's deploy script from release.sh

chrismytton:
These two scripts are very similar, so they now link to each other so
that any changes can be either kept in sync or possibly refactored into
a shared library at some point.

Related to https://github.com/everypolitician/everypolitician-data/pull/17407


chrismytton:
@tmtmtmtm 👀

PR #15556, opened 2016-09-20
Do a shallow clone of viewer-static in release.sh

chrismytton:
When deploying the site we don't need to do a full ~700MB clone of
viewer-static as we're not using the history in the repository. This
should help speed up the deploy process a bit.

Related to https://github.com/everypolitician/everypolitician-data/pull/17407 where I made a similar change to that script.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -12,7 +12,7 @@ build_viewer_static() {
 }
 
 deploy_viewer_static() {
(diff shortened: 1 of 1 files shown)


chrismytton:
@tmtmtmtm 👀

tmtmtmtm:
Woot!

PR #15558, opened 2016-09-21
Fix @head_tags warning

### LINKED ISSUE #14817: warning: instance variable @head_tags not initialized
The tests output a lot of warnings about `@head_tags` not being initialized. The cause of these should be fixed.

octopusinvitro:
I made a very tiny fix that removes the `@head_tags` warnings, after dabbling with the possibility of just making a dedicated erb template for this intermediate page (`redirect.erb`) that didn't use the main `layout.erb` page and discarding it.

The warnings were appearing when running a single test file separately, but not when running the rake tasks that run several test files. When setting the warning flags to `true` in the Rakefile, a lot of other warnings appear but those may be investigated and fixed in a different pull request.

Closes #14817

Fixup commits must be autosquashed before merging.

### DIFF
--- a/views/layout.erb
+++ b/views/layout.erb
@@ -1,5 +1,6 @@
 <% @country = @page.country if (@page and @page.respond_to? :country) %>
 <% @house   = @page.house   if (@page and @page.respond_to? :house) %>
+<% @head_tags = @head_tags ||= "" %>
 
 <!DOCTYPE html>
 <html class="no-js">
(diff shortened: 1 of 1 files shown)


views/layout.erb
```diff
@@ -1,5 +1,6 @@
 <% @country = @page.country if (@page and @page.respond_to? :country) %>
 <% @house   = @page.house   if (@page and @page.respond_to? :house) %>
+<% @head_tags = @head_tags ||= "" %>
```
tmtmtmtm: surely this should just be `@head_tags ||= ""`?

tmtmtmtm: (Though more generally, having the templates modify instance variables seems like a really bad idea to me).

PR #15559, opened 2016-09-21
Update lower redirection comment in app.rb

octopusinvitro:
The comment was pointing to a shell task that was updated to be
splitted in two separate tasks, to build and deploy the static site,
in commit 0f36ab88bbd92437fb69c6094e03c4c41014c99e

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -132,7 +132,7 @@ end
 # This works for humans (i.e., browsers parse and follow the redirect) but,
 # because wget simply fetches the HTML document, this lets us continue to
 # spider the site to generate the contents of everypolitician/viewer-static.
-# See scripts/release.sh (update_viewer_static).
+# See scripts/release.sh (build_viewer_static).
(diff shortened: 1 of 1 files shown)


app.rb
```diff
@@ -132,7 +132,7 @@
 # This works for humans (i.e., browsers parse and follow the redirect) but,
 # because wget simply fetches the HTML document, this lets us continue to
 # spider the site to generate the contents of everypolitician/viewer-static.
-# See scripts/release.sh (update_viewer_static).
+# See scripts/release.sh (build_viewer_static and deploy_viewer_static).
```
tmtmtmtm: I think we only really need to be pointing at `build_viewer_static` here, as that's the part that's really relevant to this. This used to point at the whole method because that's all there was, but now that it's split, we can just point to the `build` part.

octopusinvitro: @tmtmtmtm Addressed feedback in a fixup commit

PR #15566, opened 2016-09-23
Use tidy-html5 to validate the house page

octopusinvitro:
# What does this do?

HTML-validates the house page and removes skip duplication

# Why was this needed?

The HTML was broken and the skip was duplicated

# Relevant Issue(s)

# Implementation notes

None

# Screenshots

None

# Notes to Reviewer

Errors corrected:
- An "img" element must have an "alt" attribute, except under certain
conditions. For details, consult guidance on providing text alternatives
for images. At line 254, column 69.
- End tag "div" seen, but there were open elements. At line 256, column 12.
- Unclosed element "a". At line 252, column 77.

# Notes to Merger

None

### DIFF
--- a/t/web/house.rb
+++ b/t/web/house.rb
@@ -22,7 +22,6 @@ describe 'House Page' do
 
   describe 'HTML validation' do
     it 'has no errors in the house page' do
-      skip if `which tidy`.empty?
       last_response_must_be_valid
     end
   end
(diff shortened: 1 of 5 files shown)


t/web/house.rb
```diff
@@ -19,4 +19,11 @@
       subject.css('.avatar-unit p').last.text.must_equal '1981-01-06 - 1983-01-03'
     end
   end
+
+  describe 'HTML validation' do
+    it 'has no errors in the house page' do
+      skip if `which tidy`.empty?
```
tmtmtmtm: I don't think we want to duplicate this in every test like this. I'd suggest pulling this up into `last_response_must_be_valid`

octopusinvitro: I'm just cherry picking now, there is a commit that fixes that

PR #15571, opened 2016-09-26
Update download page URL on term table

wfdd:
This updates the download button's URL on term tables to point to the house download page rather than the late country download page.  I changed the relative to an absolute URL 'cause octopusinvitro had done the same in #15255.


tmtmtmtm:
Good catch, thanks!

Issue #15572, opened 2016-09-26
Silence test warnings


tmtmtmtm:
# Problem

Adding `t.warning = true` to each test block in the `Rakefile` generates over 20,000 warnings from a `rake test:all`

# Proposed Solution

Silence these, and have the warnings turned on by default.

# Acceptance Criteria

`rake test:all` doesn't issue any warnings with `t.warning = true` enabled.

# Not Required

# Prerequisites

# Related Issues

https://github.com/everypolitician/viewer-sinatra/issues/14817

# Due Date

# Design?

# Text?

# Bloggable?

PR #15574, opened 2016-09-26
Return objects from membership.area and group

tmtmtmtm:
We currently have two conflicting places injecting methods into Membership objects for 'area' and 'group' — MembershipExtension, which returns the relevant objects of Area or Organization; and
`person_memberships` in the TermTable page, which then overrides these to be strings. This leads to all manner of odd and inconsistent behaviour.

Change this to use objects throughout by simply falling back on the version from the MembershipExtension.

(NB: This should also remove a *lot* of the test warnings, per https://github.com/everypolitician/viewer-sinatra/issues/15572)

### DIFF
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -93,13 +93,7 @@ module Page
     end
 
     def person_memberships(person)
-      membership_lookup[person.id].each do |mem|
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
@chrismytton 👀  (cc: @ondenman)

Issue #15573, opened 2016-09-26
Test for suitable version of html-tidy


tmtmtmtm:
# Problem

The html-tidy tests fail if you have too old a version of html-tidy

# Proposed Solution

Make the skip check (which currently only tests for `which tidy` being true) more robust.

# Acceptance Criteria

Minimal acceptance is that the tests no longer fail on @tmtmtmtm's laptop!

# Not Required

# Prerequisites

# Related Issues

# Due Date

# Design?

# Text?

# Bloggable?

chrismytton:
Looks like the easiest way to do this will be to check for some expected behaviour (thanks for the suggestion @tmtmtmtm!)

The following command should produce no output and should have a `0` exit status:

```
echo '<!DOCTYPE html><title>Test</title><body><header>Test</header>' | tidy -eq
```

PR #15575, opened 2016-09-26
Check for supported version of html-tidy in tests

chrismytton:
# What does this do?

This change adds a more robust check for a supported version of html-tidy.

# Why was this needed?

Previously we were only checking that html-tidy was installed, however this meant that the user might have a very old version of html-tidy installed, which doesn't support HTML5. This new check now ensures that the user's version of html-tidy actually supports HTML5.

# Relevant Issue(s)

Closes https://github.com/everypolitician/viewer-sinatra/issues/15573

# Implementation notes

I've extracted the check for a supported version of html-tidy into it's own helper method to make it clearer what the check is actually doing.

### DIFF
--- a/t/test_helper.rb
+++ b/t/test_helper.rb
@@ -57,12 +57,19 @@ module Minitest
     end
 
     def last_response_must_be_valid
-      skip if `which tidy`.empty?
+      skip unless supported_html_tidy_version?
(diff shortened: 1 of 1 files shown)


chrismytton:
@tmtmtmtm 👀

PR #15577, opened 2016-09-26
Use tidy-html5 to validate the term-table page 

octopusinvitro:
# What does this do?

HTML-validates the term-table page and in doing so addresses [this comments](https://github.com/everypolitician/viewer-sinatra/pull/15433#discussion_r80199278)

# Why was this needed?

HTML was broken

# Relevant Issue(s)

# Implementation notes

None

# Screenshots

None

# Notes to Reviewer

Errors corrected:
- An "img" element must have an "alt" attribute, except under certain
conditions. For details, consult guidance on providing text alternatives
for images. (All occurrences)

# Notes to Merger

This branch has been rebased with master. An HTML helper is being introduced as part of https://github.com/everypolitician/viewer-sinatra/pull/15584, a helper to which this PR would be adding new methods, so that PR should probably be merged first so that adding a new helper is not part of this PR, but just modifying it.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -12,10 +12,12 @@ require 'everypolitician'
 require 'everypolitician/popolo'
 
 require_relative './lib/popolo_helper'
+require_relative './lib/html_helper'
 require_rel './lib/page'
 
 Dotenv.load
 helpers Popolo::Helper
+helpers HTMLHelper
 
 set :erb, trim: '-'
 set :main_url, 'http://everypolitician.org'
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -154,10 +154,13 @@
                                 <img src="/images/person-placeholder-108px.png"
                                      data-src="<%= person.proxy_image %>"
                                      style="display: none"
-                                     class="person-card__image">
-                                <noscript><img src="<%= person.proxy_image %>" class="person-card__image"></noscript>
+                                     class="person-card__image"
+                                     alt="Member headshot for <%= escape_html(person.name) %>">
+                                <noscript><img src="<%= person.proxy_image %>" class="person-card__image"
+                                alt="Member headshot for <%= escape_html(person.name) %>"></noscript>
                               <% else %>
-                                <img src="/images/person-placeholder-108px.png" class="person-card__image">
+                                <img src="/images/person-placeholder-108px.png" class="person-card__image"
+                                alt="Placeholder image for <%= escape_html(person.name) %>">
                               <% end %>
 
                                 <h3 class="person-card__name"><%= person.name %></h3>
@@ -211,7 +214,7 @@
       <div class="page-section page-section--grey source-credits">
         <div class="container">
           <% if @page.data_sources %>
-            <p>Main Source<% if @page.data_sources.size > 1 %>s<% end %>: <%= @page.data_sources.map { |url| %Q(<a href="#{url}">#{url}</a>) }.join(", ") %></p>
+            <p>Main Source<% if @page.data_sources.size > 1 %>s<% end %>: <%= @page.data_sources.map { |url| %Q(<a href="#{escape_uri(url)}">#{url}</a>) }.join(", ") %></p>
           <% end %>
           <p><b>Anything wrong?</b> If you've spotted an error, or the data is incomplete,
             here's <a href="http://docs.everypolitician.org/contribute.html">how to get that fixed</a>.</p>
--- a/lib/html_helper.rb
+++ b/lib/html_helper.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+module HTMLHelper
+  def escape_html(text)
+    Rack::Utils.escape_html(text)
+  end
+
+  def escape_uri(text)
+    URI.escape(text)
+  end
+end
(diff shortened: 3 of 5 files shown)


t/web/term_table/phillipines.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../../app'
+
+describe 'Phillipines' do
```
tmtmtmtm: Typo: "Philippines"

app.rb
```diff
@@ -12,6 +12,7 @@
 require 'everypolitician/popolo'
 
 require_relative './lib/popolo_helper'
+require_relative './helpers/html'
```
tmtmtmtm: Shouldn't we be consistent where the helpers live? If we're making a new 'helpers' directory for them, then we should probably move the `popolo_helper` there too.

helpers/html.rb
```diff
@@ -0,0 +1,12 @@
...
+require 'sinatra/base'
+
+module Sinatra
+  module HTMLEscapeHelper
+    def h(text)
+      Rack::Utils.escape_html(text)
+    end
+  end
+
+  helpers HTMLEscapeHelper
```
tmtmtmtm: With the `PopoloHelper` we [set it up](https://github.com/everypolitician/viewer-sinatra/blob/ee9d1dd664f4094a85e9ba7b38338213fb14b570/app.rb#L18) in `app.rb` itself. I'm not sure whether it's generally preferred to do it that way or this, but again I think we should be consistent.

t/web/term_table/phillipines.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    it 'has the person name in a placeholder image' do
+      img = dimaporo.css('img.person-card__image')
+      img.attr('alt').text.must_equal 'Placeholder image for Dimaporo, Abdullah D.'
+    end
+
+    it 'doesnt break for names with double quotes' do
+      img = tiangco.css('img.person-card__image')
+      img.attr('alt').text.must_equal 'Member headshot for Tiangco, Tobias "Toby" M.'
```
tmtmtmtm: There are actually two different images with this text, and only one of them is getting checked here. Both of them actually get returned from this css selector, though, so it should be quite easy to check both.

helpers/html.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+require 'sinatra/base'
+
+module Sinatra
+  module HTMLEscapeHelper
+    def h(text)
```
tmtmtmtm: Is `h` a general standard for naming functions like this? I'm generally a fan of brevity, but I'm a little hesitant here, unless this is a universally recognised approach.

app.rb
octopusinvitro: I thought about that and I think it may be a good idea. I've seen that in other frameworks as Rails or Middleman and I think it helps to keep concerns separated. But I am happy to put this file under lib if we prefer it so.

helpers/html.rb
octopusinvitro: Ah, this was copied from [the official documentation](http://www.sinatrarb.com/extensions.html#extending-the-request-context-with-sinatrahelpers), I can certainly rename it

helpers/html.rb
octopusinvitro: See previous comment. I am happy to move it to app.rb though.

octopusinvitro:
@tmtmtmtm Feedback addressed in fixup commits

* I tried to call the HTML helper from the app but it couldn't make it work. So I just moved the helper from the helper folder to the lib folder. I also renamed the `h` method to `escape_html`.
* I downloaded and used the popolo file from the Bahamas and removed the one for the Philippines.
* I removed the test file for the Philippines and added a new one for the Bahamas. The switch to the Bahamas file allowed me to discover another bug, when one of the urls in the sources of the data contains spaces. I fixed that by adding another method to the `HTMLEscapeHelper` called `escape_uri` which uses `URI.escape`.

views/term_table.erb
```diff
@@ -198,7 +198,7 @@
       <div class="page-section page-section--grey source-credits">
         <div class="container">
           <% if @page.data_sources %>
-            <p>Main Source<% if @page.data_sources.size > 1 %>s<% end %>: <%= @page.data_sources.map { |url| %Q(<a href="#{url}">#{url}</a>) }.join(", ") %></p>
+            <p>Main Source<% if @page.data_sources.size > 1 %>s<% end %>: <%= @page.data_sources.map { |url| %Q(<a href="#{escape_uri(url)}">#{url}</a>) }.join(", ") %></p>
```
tmtmtmtm: This approach concerns me. If the problem is that EP is giving us invalid URLs, then we should fix that upstream, so that everyone consuming the data doesn't need to do things like this. 

Whichever approach we take, however, this feels like something that deserves a PR of its own, properly explaining the problem and solution etc., rather than cramming more into _this_ PR.

tmtmtmtm:
@octopusinvitro when you say "it couldn't make it work", what was the problem?

octopusinvitro:
@tmtmtmtm This branch has been rebased with master. An HTML helper is being introduced as part of https://github.com/everypolitician/viewer-sinatra/pull/15584, a helper to which this PR would be adding new methods, so that PR should probably be merged first so that adding a new helper is not part of this PR, but just modifying it.

octopusinvitro:
Closed as it was moved to #15587

PR #15578, opened 2016-09-26
Rake default

octopusinvitro:
# What does this do?

It adds a new default task to Travis, so that running `bundle exec rake` runs everything and all the things: tests, syntax checks and security checks

# Why was this needed?

Because it was easy to forget to manually run all the things before pushing, and get different results from Travis. Now we only have to run a single command.

# Relevant Issue(s)

"_Use default rake tasks to configure what runs on Travis_" https://github.com/everypolitician/everypolitician/issues/516

# Implementation notes

None

# Screenshots

None

# Notes to Reviewer

None

# Notes to Merger

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,13 @@ bundle exec rake test:extensions
 bundle exec rake test:all
 ```
 
-It is recommended to run the last command at least once right before pushing.
+### Run absolutely all the tests plus rubocop and bundle audit
+
+```bash
+bundle exec rake
+```
+
+It is recommended to run the last command at least once right before pushing. Also, running this command assures you that what you get locally is the same that you will get in Travis.
 
 ## Sinatra, SASS, styling
 
--- a/README.md
+++ b/README.md
@@ -110,7 +110,6 @@ bundle exec rake test:all
 bundle exec rake
 ```
 
-It is recommended to run the last command at least once right before pushing. Also, running this command assures you that what you get locally is the same that you will get in Travis.
 
 ## Sinatra, SASS, styling
 
--- a/Rakefile
+++ b/Rakefile
@@ -54,3 +54,5 @@ end
 
 require 'bundler/audit/task'
 Bundler::Audit::Task.new
+
+task default: ['test:all', 'rubocop', 'bundle:audit']
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,7 @@ addons:
 before_install:
   - scripts/install_tidy.sh
 script:
-  - bundle exec rake test:all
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -104,7 +104,13 @@ bundle exec rake test:extensions
...
 ```
 
-It is recommended to run the last command at least once right before pushing.
+### Run absolutely all the tests plus rubocop and bundle audit
+
+```bash
+bundle exec rake
+```
+
+It is recommended to run the last command at least once right before pushing. Also, running this command assures you that what you get locally is the same that you will get in Travis.
```
tmtmtmtm: This paragraph is a little odd. Travis is introduced with no prior explanation, or with any indication of why it's important, and with the implication that getting something from Travis is a goal. But this all seems slightly back-to-front. Travis merely checks that all these things pass — **it** is not really the goal by itself. 

If we're going to spell this all out, then I think we should do it even more explicitly, explaining what checks we run{{1}}, and explaining that we have Travis set up to run those checks automatically, and that _you_, dear contributor, can run exactly the same checks through a simple `bundle exec rake`.

---

{{1}} ideally in a semi-standardised way that we can drop in to multiple projects

tmtmtmtm: PS, I realise that this is in the pre-existing version but suggesting that someone runs this _at least_ once, right before pushing, is also a bit strange. Why, right before pushing, would they run it twice or more?

octopusinvitro:
@tmtmtmtm Since the copy in that paragraph was confusing and there is not a defined standardised way to explain  this process yet, I opted for removing the paragraph completely until we find a better copy to use there.

Rakefile
```diff
@@ -54,3 +54,5 @@ end
 
 require 'bundler/audit/task'
 Bundler::Audit::Task.new
+
+task default: ['test:all', 'rubocop', 'bundle:audit']
```
chrismytton: This code is absolutely fine, but quick tip: I tend to write this using the [array of symbols shorthand](https://ruby-doc.org/core-2.3.1/doc/syntax/literals_rdoc.html#label-Percent+Strings):

```ruby
task default: %i(test:all rubocop bundle:audit)
```

✨

README.md
```diff
@@ -104,7 +104,12 @@ bundle exec rake test:extensions
 bundle exec rake test:all
 ```
 
-It is recommended to run the last command at least once right before pushing.
+### Run absolutely all the tests plus rubocop and bundle audit
```
tmtmtmtm: I'm a little concerned that this line will drift out of sync as we add other default tasks, but I can live with that for now.

PR #15580, opened 2016-09-26
Use tidy-html5 to validate the home page

### LINKED ISSUE #15562: MAke the hint text visible again
# Problem

After adding HTML validation to viewer-sinatra, some attributes had to be removed in order for the validation to pass. The “Search by country name” hint text now is missing from the search input on the homepage as a result of this.

Before:

screen shot 2016-09-20 at 17 24 35

After:

screen shot 2016-09-20 at 17 24 45

(Images by @zarino )

# Proposed Solution

> Since the text input (generated by [selectToAutocomplete](http://baymard.com/labs/country-selector)) uses `copy-attributes-to-text-field` to soak up whatever HTML attributes the source `<select>` element had. I guess, a satisfactory solution if we still want W3C compliant markup, would be to add the `placeholder` attribute back in with JavaScript, immediately before we call `. selectToAutocomplete()` in [main.js](https://github.com/everypolitician/viewer-sinatra/blob/75f3218b87e41d26e97b6ad17550630f2af5bdae/public/javascript/main.js#L308).
-- Zarino.

# Acceptance Criteria

The hint text is visible again in the search input on the homepage.

octopusinvitro:
# What does this do?

HTML-validates the home page

# Why was this needed?

HTML was broken

# Relevant Issue(s)

Closes #15562

# Implementation notes

The attributes removed to make the validation pass where added back with JavaScript.

# Screenshots

screen shot 2016-09-20 at 17 24 35

(Image by @zarino )

# Notes to Reviewer

Errors corrected:
- Attribute "placeholder" not allowed on element "select" at this point. At line 86, column 189
- Attribute "autocorrect" not allowed on element "select" at this point. At line 86, column 189
- Attribute "autocomplete" not allowed on element "select" at this point. At line 86, column 189

# Notes to Merger

None

### DIFF
--- a/public/javascript/main.js
+++ b/public/javascript/main.js
@@ -305,15 +305,23 @@ $(function(){
   });
 
   // http://baymard.com/labs/country-selector
-  $('.js-select-to-autocomplete').selectToAutocomplete().on('change', function(){
-    var v = $(this).val();
-    if (v) {
-      // Assumes the <option>'s `value` attribute is a URL slug for the country
-      window.location.href = '/' + v + '/';
-    }
-  }).on('focus', function(){
-    $(this).next().trigger("focus");
-  });
+  $('.js-select-to-autocomplete')
+    .attr({
+      placeholder:  'Search by country name',
+      autocorrect:  'off',
+      autocomplete: 'off'
+    })
+    .selectToAutocomplete()
+    .on('change', function(){
+      var v = $(this).val();
+      if (v) {
+        // Assumes the <option>'s `value` attribute is a URL slug for the country
+        window.location.href = '/' + v + '/';
+      }
+    })
+    .on('focus', function(){
+      $(this).next().trigger("focus");
+    });
 
   // Fix the incorrect default autocomplete width, which meant that
   // autocomplete menu was longer than the search input it's linked to.
--- a/t/web/basic.rb
+++ b/t/web/basic.rb
@@ -66,4 +66,12 @@ describe 'Viewer' do
       last_response.status.must_equal 404
     end
   end
+
+  describe 'HTML validation' do
+    before { get '/' }
+
+    it 'has no errors in the home page' do
+      last_response_must_be_valid
+    end
+  end
 end
--- a/views/country_selector.erb
+++ b/views/country_selector.erb
@@ -1,4 +1,4 @@
-<select class="field js-select-to-autocomplete" placeholder="Search by country name" name="Country" id="country-selector" autofocus="autofocus" autocorrect="off" autocomplete="off">
+<select class="field js-select-to-autocomplete" name="Country" id="country-selector" autofocus="autofocus">
     <option value="" selected="selected">Select Country</option>
   <% @page.all_countries.each do |country| %>
     <option value="<%= country.slug %>" data-alternative-spellings="<%= country.names %>" title="<%= commify(country.total_people) %> <% if country.total_people == 1 %>person<% else %>people<% end %>"><%= country.name %></option>


octopusinvitro:
@tmtmtmtm I added a screenshot.

zarino:
> (Image by @zarino )

Maybe I'm missing something, but what's the point of including a screenshot in a PR, if the screenshot was taken *before* the changes in the PR were even written?

octopusinvitro:
@zarino Because nothing has changed and I am too lazy to take another identical screenshot! :smile:

davewhiteland:
I would like to be convinced that we need to do this: `placeholder` works as an inbuilt feature of modern browsers, without any requirement for JS (yay). But this way we're only using `placeholder` if JS is enabled/available, which seems like an accessibility step backwards.

My understanding of this is that we're using JS to add these attributes after validation because the validator will otherwise report them. If so, it _feels_ odd to be introducing a JS dependency to get around a problem with the validator.

zarino:
> My understanding of this is that we're using JS to add these attributes after validation because the validator will otherwise report them. If so, it feels odd to be introducing a JS dependency to get around a problem with the validator.

We’re adding them because, to the letter of the law, the HTML5 spec doesn't allow a `placeholder` attribute to be applied to a `select` element.

If you ask me – it's a classic example of when you safely ignore the spec. But that's just me.

octopusinvitro:
@davewhiteland The easy fix for that would be to include actual DOM element in the document just before the select with the placeholder/informative, text, and if JS is enabled, it would hide that element, pick the informative text it contains, and use it as a placeholder in the select.

@zarino what do you think?

zarino:
> @zarino what do you think?

Including the placeholder text in a `<label>` or whatever, before the `<select>`, and then removing the element and applying its content as the `placeholder` attribute on the `<select>` before it gets `selectToAutocomplete`’d—as you suggest—is probably the best idea. Not only does it gracefully degrade if JavaScript doesn't run, but it also avoids the thorny semantic argument over whether a command like “Search by country name” is actually a suitable thing to be put into a `placeholder` attribute.

Question is: what’s the best way of marking up such a “hint” element. A second label is one option:

```erb
<p>
  <label for="country-selector">Find representatives from your country:</label>
  <label for="country-selector">Search by country name</label>
</p>
<%= erb :country_selector %>
```

A separate element inside the same label is another option:

```erb
<p>
  <label for="country-selector">
    Find representatives from your country:
    <span class="js-turn-me-into-a-placeholder">Search by country name</span>
  </label>
</p>
<%= erb :country_selector %>
```

octopusinvitro:
@zarino I like the second one better. 😃

However I would add a third option: The text "Search by cou

octopusinvitro:
@zarino I like the second one better 😃

 I would add a third one:
The text "Search by country name" feels more like a proper label for the select. So another option could be to make that be the actual label and "Find representatives in your country" can be a heading or another paragraph:

```erb
<p class="fancy-paragraph">Find representatives from your country:</p>
<p class="turn-me-into-a-placeholder">
  <label for="country-selector">Search by country name</label>
</p>
<%= erb :country_selector %>
```

zarino:
> The text "Search by country name" feels more like a proper label for the select. So another option could be to make that be the actual label and "Find representatives in your country" can be a heading or another paragraph

I'd say that's fine, as long as you can still click the "Find representatives in your country" and it focusses the search box. (That’s something you get for free when you use a `<label>`.)

octopusinvitro:
@zarino ah yes you are right. I'll go with your second option :+1:

tmtmtmtm:
I think I'm going to put a bit of kibosh on this one, at least for now. I can certainly see _some_ value in having all our HTML be completely valid, but this is already several levels removed from the original trigger for this, which was about detecting mismatched tags. 

The value of the changes being discussed here now seem a little too minimal for the amount of effort, when we have so many other important things to do.

PR #15581, opened 2016-09-27
Remove duplicate before_install section in .travis.yml

chrismytton:
# What does this do?

This removes a duplicate `before_install` section from `.travis.yml` that I accidentally added in #15573.

# Why was this needed?

Deploys of everypolitician.org were failing because I'd overwritten the `before_install` section that was adding credentials to `.netrc`, so git didn't have the correct credentials available to push to viewer-static.

# Relevant Issue(s)

#15573 was the pull request that introduced the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,11 @@ language: ruby
 cache: bundler
 before_install:
   - echo -e "machine github.com\n  login $GITHUB_ACCESS_TOKEN" >> ~/.netrc
+  - scripts/install_tidy.sh
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


chrismytton:
@tmtmtmtm :eyes:

chrismytton:
A little digging around led me to https://github.com/ruby/psych/issues/79#issuecomment-228477830 which suggests using https://github.com/adrienverge/yamllint. That's written in python but _should_ be possible to get it running on Travis (it would be nice if Travis could spit out a warning but oh well!)

PR #15582, opened 2016-09-27
Document that person image can show if JS is disabled

octopusinvitro:
# What does this do?

It improves the image tests in the term-table page tests for Northern Ireland, documenting the fact that a person with an image shows it both when JS is and is not disabled.

It also documents that when there is no image available for a person, a placeholder image will be shown.
# Why was this needed?

Because there were no tests for it.
# Relevant Issue(s)

None
# Implementation notes

None
# Screenshots

None
# Notes to Reviewer

None
# Notes to Merger

None

### DIFF
--- a/t/web/term_table/ni.rb
+++ b/t/web/term_table/ni.rb
@@ -31,9 +31,21 @@ describe 'Northern Ireland' do
     end
 
     it 'should start with a placeholder image, but have proxy set' do
-      img = annalo.css('img.person-card__image')
+      img = annalo.css('img.person-card__image')[0]
+      img.attr('src').must_equal '/images/person-placeholder-108px.png'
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
@octopusinvitro I'm a little confused by this one. The description says that it's about _documenting_ things, but the only thing that has changed is tests. If the issue is that we need documentation, then this doesn't seem to be addressing that. If the issue is simply that we're missing _tests_, then I think the description should reflect that.

octopusinvitro:
@tmtmtmtm ah, sorry, is a dialectic thing. Tests are indeed documentation of how to use your code and how it works, the features it has, etc. That doesn't mean that there aren't other types of documentation as well.

Issue #15583, opened 2016-09-29
The URLs are been unencoded  before being passed to the template


octopusinvitro:
# Problem

In https://github.com/everypolitician/viewer-sinatra/commit/25e4f5b479113fb7010fceded92923e8772f4fc8, a change was introduced where the URLs were unencoded before being passed to the template, producing Bad Urls for linking (although totally fine for displaying).

This change has survived to this point and now lives in the [term table page](https://github.com/everypolitician/viewer-sinatra/blob/master/lib/page/term_table.rb#L18), created when moving the logic out of the routes.

# Proposed Solution

The URLs should be unencoded at the template level, using for example a helper. The url should arrive to the template as a valid URL so that it can be used as a valid URL to link to.

# Acceptance Criteria

The HTML validator passes (link URL is valid) and the URLs are displayed unencoded (displayed URL is human readable).

PR #15584, opened 2016-09-29
Unescape urls

octopusinvitro:
# What does this do?

Unescapes the URLs of the sources at the template level.

# Why was this needed?

We were unescaping the URLs at the page level, so we were sending invalid urls to the template that we couldn't link to and were making the HTML validator complain.

# Relevant Issue(s)

Closes https://github.com/everypolitician/viewer-sinatra/issues/15583

# Implementation notes

A new helper file was added and the unescaping was moved from the term-table page to the term-table template, using the new helper

# Screenshots

Displayed urls

urls

Linked urls

code

# Notes to Reviewer

None

# Notes to Merger

After this PR was published and reviewed, fixup commits were added that need auto-squashing before merging.

### DIFF
--- a/lib/html_helper.rb
+++ b/lib/html_helper.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+require 'sinatra/base'
+
+module Sinatra
+  module HTMLHelper
+    def unescape_uri(text)
+      CGI.unescape(text)
+    end
+  end
+end
--- a/lib/html_helper.rb
+++ b/lib/html_helper.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
-require 'sinatra/base'
-
-module Sinatra
-  module HTMLHelper
-    def unescape_uri(text)
-      CGI.unescape(text)
-    end
+module HTMLHelper
+  def unescape_uri(text)
+    CGI.unescape(text)
   end
 end
--- a/t/web/term_table/bahamas.rb
+++ b/t/web/term_table/bahamas.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+require 'test_helper'
+require_relative '../../../app'
+
+describe 'Bahamas' do
+  subject { Nokogiri::HTML(last_response.body) }
+
+  before do
+    stub_popolo('4da60b8', 'Bahamas/House_of_Assembly')
+    get '/bahamas/house-of-assembly/term-table/2012.html'
+  end
+
+  describe 'source urls' do
+    it 'links to a valid url' do
+      subject.css('.source-credits p:first-child a').last.text.must_include '/Members of Parliament/'
+    end
+
+    it 'displays an unescaped url' do
+      subject.css('.source-credits p:first-child a/@href').last.text.must_include '/Members+of+Parliament/'
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


t/web/term_table/bahamas.rb
```diff
@@ -0,0 +1,22 @@
...
+  subject      { Nokogiri::HTML(last_response.body) }
+
+  before do
+    stub_popolo('4da60b8', 'Bahamas/House_of_Assembly')
+    get '/bahamas/house-of-assembly/term-table/2012.html'
+  end
+
+  describe 'source urls' do
+    it 'links to a valid url' do
+      subject.css('.source-credits p:first-child a').last.text.must_include '/Members of Parliament/'
```
tmtmtmtm: as both tests are operating on essentially the same element, perhaps that would be better as the `subject`?

octopusinvitro:
@tmtmtmtm feedback addressed

lib/html_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+require 'sinatra/base'
```
tmtmtmtm: Why is this required? I can't see how this helper is really connected with Sinatra at all, and if I remove this the tests all pass, and the Bahamas page still seems to do the right thing.

lib/html_helper.rb
```diff
@@ -0,0 +1,10 @@
...
+
+module Sinatra
```
tmtmtmtm: Connected to the above point, but slightly separate from it, what's your thinking behind adding this to the Sinatra namespace? It seems a little odd to me, though I suspect that's just because I'm missing something.

octopusinvitro: This helper was taken from the [official Sinatra documentation](http://www.sinatrarb.com/extensions.html#extending-the-request-context-with-sinatrahelpers) on how to add an HTML escape helper

octopusinvitro:
@tmtmtmtm feedback addressed

lib/html_helper.rb
tmtmtmtm: I think you might have misunderstood my comment. I wasn't saying "you should change this". I was asking why it was like this. In general it's good to avoid simply cargo-culting code, even from official documentation. It's better to understand _why_ an example is a certain way. Is it's because it's the way you *need* to do something? (In this case that seems unlikely if it still works now that you've done it differently.)  With something like Sinatra, with two main modes of operation, perhaps an example is particularly geared towards one or the other? Or perhaps it's a convention that, whilst not strictly necessary, would be good to follow in the absence of any compelling reason why _not_ to. Or perhaps it's just one of many possible ways to achieve the same thing, and we're better sticking with something that more fits our existing codebase and approach. (We already have another Helper file. Introducing a second that works differently makes it a little more confusing for anyone reading the code, and harder for someone to know what to do when introducing another one. The two files don't necessarily need to be perfectly harmonised, but if they're different, we should be doing so knowingly.)

NB: when you _do_ take code from elsewhere, you should credit that.

octopusinvitro: Great comment, thank you!
Let me investigate that. :+1:

octopusinvitro:
Closing as it has been moved to #15588

PR #15588, opened 2016-10-14
Unescape URLs of Sources

octopusinvitro:
# What does this do?

Unescapes the URLs of the sources at the template level.

# Why was this needed?

We were unescaping the URLs at the page level, so we were sending invalid urls to the template that we couldn't link to and were making the HTML validator complain.

# Relevant Issue(s)

Closes https://github.com/everypolitician/viewer-sinatra/issues/15583

# Implementation notes

A new helper file was added and the unescaping was moved from the term-table page to the term-table template, using the new helper.

It also adds a new fixture for the Bahamas which is an example of the urls this PR fixes.

# Screenshots

Displayed urls

urls

Linked urls

code

# Notes to Reviewer

None

# Notes to Merger

~~After this PR was published and reviewed, fixup commits were added that need auto-squashing before merging.~~

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -12,10 +12,12 @@ require 'everypolitician'
 require 'everypolitician/popolo'
 
 require_relative './lib/popolo_helper'
+require_relative './lib/html_helper'
 require_rel './lib/page'
 
 Dotenv.load
 helpers Popolo::Helper
+helpers HTMLHelper
 
 set :erb, trim: '-'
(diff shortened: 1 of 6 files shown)


tmtmtmtm:
@octopusinvitro the Notes to Merger claim "After this PR was published and reviewed, fixup commits were added that need auto-squashing before merging.", but I can't see any fixups here at all…

Issue #15589, opened 2016-10-24
Ask people what data we should gather for each country


tmtmtmtm:
# Problem

We want to add a lot more data (and not  about who politicians are, but also what they do). But there's a _lot_ of data to gather, for a _lot_ of countries, and we need to find some way to prioritise.
# Proposed Solution

Ask visitors to the site what they'd like to see us add (ideally because they'd actually like to **use** it)

For example, when viewing a page for a country, we ask them "Would you like to see more data for this country/legislature?"

Capture, somewhere, which ones get 'voted' for like this.

Preferably, we could also let them sign up to be notified when we add more.

Ideally, we'd let them say what types of data they'd like us to prioritise importing: possibilities include biographical information, contact details, social media profiles, political history, education history, voting records, attendance records, committee memberships, speeches, bill authorship/sponsorship, promises made, income + asset declarations, etc…)

As well as gaining lots of potentially useful information, this would also help signal that we're actively adding new data, and want to be much more than what we currently have.
# Acceptance Criteria

Minimal version: we get a list of 'votes' for each country/legislature.

Preferable: we get individual results (with optional email addresses)

Ideal: We get detail on what data they want, and why.
# Not Required

We don't need to store/gather this information in any specific way, at least not at first. If we happen to start getting LOTS of people telling us what they want, then we might want to look at some structured way of storing it, but let's leave that until we need it.
# Prerequisites
# Related Issues
# Due Date

Ideally before Mozfest.
# Design?

Preferably!
# Text?
# Bloggable?

zarino:
@tmtmtmtm Would you be happy to do this via Google Forms for now? So, something like:

* Banner on legislature pages, encouraging people to **click here** if they want us to collect more data for this legislature.
* Button in banner leads to the Google Form, hopefully styled in EveryPolitician colours so it doesn’t look too alien.
* Responses get saved to a Google Sheet.
* Respondent is sent back to everypolitician.org (or maybe we just open the Google Form in a new tab?)

tmtmtmtm:
@zarino: sure, that sort of thing could work well.

JenMysoc:
If you also restrict what answers people can give (eg don't let them just write free text) then we can get graphs to show us the most popular countries/types of data

zarino:
[...]

PR #15590, opened 2016-10-25
New section for survey on legislature page

zarino:
# What does this do?

Displays a bright yellow section between the “What’s in this dataset?” and “Members” sections of the legislature / term_table page, advertising the upcoming Google Form for collecting data requests.

# Why was this needed?

Because we want to find out what data people want, and about which legislatures.

# Relevant Issue(s)

#15589 

# Implementation notes

~~We’ll need to set the orange button to link to a Google Form, with the page URL (or maybe the country name and legislature name?) passed into the form as [pre-populated answers](https://support.google.com/docs/answer/160000?hl=en).~~ DONE

# Screenshots

screen shot 2016-10-28 at 12 22 48

# Notes to Reviewer

Not ready to review yet! Wording not final. Link doesn’t go anywhere.

# Notes to Merger


zarino:
@tmtmtmtm – I’d appreciate your feedback on whether this is too prominent, or not prominent enough. And whether the wording is heading in the right direction.

tmtmtmtm:
@zarino the main thing that this reveals is that we might want two separate processes — one for "more data about these politicians" and one for "more data from this country". One of the things we definitely need to find out is what other data people want — e.g. in the UK are there people who really want data further back 1997, or in the US do they want data about historic presidents and cabinet members (who aren't in the legislature), etc. Not sure whether we can do that all in one shot, but perhaps having parallel banners like this on the country / legislature pages could work well. 

Otherwise, this is definitely along the right sort of lines.

zarino:
@tmtmtmtm – I wonder whether two separate forms is just over-complicating matters. Could we get away with just one form, and one banner (on the term_table page), if we changed the wording?

zarino:
@tmtmtmtm – It sounds like you’re not too worried about gathering structured data right now, so how about keeping the banner and the form really really open-ended?

The banner could be worded like:

> **Do you need more data?**
> EveryPolitician is growing fast! Help us prioritise by letting us know what data would be most useful to you.
> [ Place your vote now ]

And the form could basically just ask:

> **Legislature** (pre-filled by Google Forms based on referring page)
> [ http://everypolitician.org/…html ]
>
> **What data do you need, that we don’t currently have?**
> [ ________________ ]
>
> **Enter your email address if you’d like to be notified when we add more data**
> [ ________________ ]

tmtmtmtm:
@zarino, you're right that we don't need structured data, but my inclination is to make it a little less open-ended than this. That might just be about some explanation as to the sorts of data we're looking at, rather than actively letting people select between those, though.

(I'd also soften "do you need" to something like "could you use", and for the email address question, I think we need to be clear as to whether this is "we'll personally let you know when we add the data you're asking for" vs "you're joining a mailing list with info on all updates everywhere")

JenMysoc:
I think we do actually need two separate sections to the form. I don't think it's complicating it if we make it clear what the difference is as @tmtmtmtm said above. Then we can quickly see ourselves who wants what. We can actually split the sections on the form itself though. So we could have an explanation like "Could you use more information about politicians and their activities in this country" except better worded... 

Then when people get to the form it asks them one question: 
Do you want more information about politicians for this legislature (eg Biographical details, education details etc) 
OR
Do you want more details about this legislature (eg historical data, term data etc)

Then the answer to that question could take them to a different page based on what answer they give.

That page could then ask two questions:
What data are you looking for?
_they can choose from multiple choice boxes - we can set a limit if you want on how many they can choose and we can add a section for other if you want as well!_

Would you provide your email address so we can send an update when we have this data?

davewhiteland:
We should ask anyone willing to fill a form in _why_ they want the data, if they know (that is, an optional (text) input); that is, it's really useful to know the sort of things people want this data for (especially if it's counter to our own assumptions).

Also be cause about the "more detail or more depth" either-or question because it seems very reasonable that someone might want both, no?

davewhiteland:
More possibilities:

* sign up to mySociety newsletter? y/n

* sign up to EveryPolitician newsletter? y/n  *if we are willing to maintain a newsletter*

* (here's the magic) want us to send you an email automatically whenever data in the legislature you're interested in changes? y/n (TODO: wee task that we run that sends emails out, which is triggered by a webhook, innit?)

MyfanwyNixon:
"Place your vote now" seems confusing in context.

Could change to:
>EveryPolitician is growing fast! Help us prioritise: what data would be useful for you?
> [button text] Tell us now

zarino:
@MyfanwyNixon – Changed! (See screenshot at the top of this page)

MyfanwyNixon:
👍

chrismytton:
Following on from a Slack discussion with @tmtmtmtm I've just added the banner to the country and house page. I've also removed the name of the house from the survey url to make things read a bit better on the form.

PR #15591, opened 2016-10-26
More mobile-friendly header and nav menu

zarino:
# What does this do?

The old `.site-header` is replaced with a `.site-nav` and, on country pages only, an additional `.site-header` to hold the country name and legislature name, if available.

It also reduces the amount of space taken up by the header/nav on narrow screens, and uses javascript to hide the nav links away behind a "Menu" button on narrow screens.

# Why was this needed?

On narrow screens, some people were mistaking the "E" of the EveryPolitician logo for a hamburger menu.

We can’t change the logo, but we *can* give them a clearly labelled "Menu" button to press instead.

The logo is made smaller, and is no longer centred, further avoiding the illusion that "EveryPolitician" was some sort of app-like header bar that could be pressed in its entirety.

# Relevant Issue(s)

#11912 

# Implementation notes

# Screenshots

## Narrow screen

`.site-nav` on homepage (showing menu reveal interaction):

dropdown

`.site-nav` and `.site-header` on a legislature page:

legislature mobile

## Medium screen

`.site-nav` on homepage:

screen shot 2016-10-26 at 15 57 37

`.site-nav` and `.site-header` on a legislature page:

screen shot 2016-10-26 at 15 56 13

## Large screen

`.site-nav` on homepage:

screen shot 2016-10-26 at 15 58 29

`.site-nav` and `.site-header` on a legislature page:

screen shot 2016-10-26 at 15 58 51

# Notes to Reviewer

# Notes to Merger


zarino:
@tmtmtmtm Does this look ok to you?

@davewhiteland – As a live witness of a user making the "clicking the logo" mistake – do you want to try this out on your own mobile device and let me know whether you think it’ll help.

Worth saying, there’s one final extra step we can take, if we find people *still* try to click the logo to trigger the menu – we can just make clicks on the logo *actually* open the menu (in addition to the existing "Menu" button) and then add a "Home" link to the menu, to let people get back to the homepage if they want to go there. The extra "Home" link in the nav, and the logo click handler, would only be there on narrow screens (ie: when there is a hidden menu to be toggled).

tmtmtmtm:
If it gets the @davewhiteland :+1:, then it's 👍  from me too :)

Thanks @zarino

PR #15601, opened 2017-01-03
Show politicians’ names as they were at term end

andylolz:
Refs everypolitician/everypolitician/issues/384.

Apologies if this is nonsense. My ruby is bad.

### DIFF
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -59,6 +59,7 @@ module Page
       @people ||= people_for_current_term.sort_by(&:sort_name).map do |person|
         PersonCard.new(
           person:          person,
+          person_name:     person.name_at(term.end_date),
           proxy_image:     image_proxy_url(person.id),
           memberships:     person_memberships(person),
           top_identifiers: top_identifiers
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
Hi @andylolz — thanks for this, and sorry for the delay in getting around to it. In terms of implementation, my main concern is that it's adding another argument to a constructor that already has a comment of "# TODO: pass fewer arguments!" — but I can look separately at how to tidy that up. 

My wider issue here, though, is that I thought https://github.com/everypolitician/everypolitician-popolo/issues/119 was saying that this would have no effect? I'm assuming now I've misunderstood something here, but either way I think there'd need to be a test case here that fails against the current `master` — i.e. that shows some name changing when implemented. I'm happy to add that test case myself if you're unable to, if you can point me in the direction of a suitable example.

andylolz:
Hi @tmtmtmtm, happy new year! Yes, you've understood correctly! The underlying problem remains. I've added https://github.com/everypolitician/everypolitician/issues/384#issuecomment-270109836 for clarification on that.

In the meantime, this PR is to ensure the end of term name is used once data problems sorted. Good point re. test case! I'll have a go at that.

andylolz:
Looks like this was with me to add tests, but I’m afraid I didn’t got round to it. Apologies!

In any case, it’s doubtful it’s still applicable.

PR #15603, opened 2017-01-17
Simplify the instantiation of PersonCard objects

mhl:
# What does this do?

This pull request simplifies the dependencies of the `PersonCard` class,
so that it only needs to take a `Person` and a `LegislativePeriod` when
instantiating. It also moves some functionality from `TermTable` to
`LegislativePeriod` ~~(by monkey-patching the latter class).~~

# Why was this needed?

In the future we would like to reuse these models in other projects, and
these changes make that reuse easier.

# Relevant Issue(s)

I couldn't find one, but some of these changes were suggested by "TODO"
comments in the existing source.

# Implementation notes

~~I'm not 100% sure if the monkey-patching of `LegislativePeriod` is sensible
or not - maybe similar changes in the original class in `everypolitician-ruby`
would be more useful in the long run, but I thought this was OK.~~

# Screenshots

The appearance of the site should be unchanged by these.

# Notes to Reviewer

I've run the recursive `wget` on the site before and after this change, and
the results are identical (according to `diff -r`) so I don't think this has
broken anything.

~~I'm not sure if the upgrade of everypolitician-popolo is actually required
any more (I've gone through several variants of this change) but it seems
like a good idea in any case.~~

# Notes to Merger

None.

### DIFF
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -59,7 +59,6 @@ module Page
       @people ||= people_for_current_term.sort_by { |e| [e.sort_name, e.name] }.map do |person|
         PersonCard.new(
           person:          person,
-          proxy_image:     image_proxy_url(person.id),
           term:            term,
           top_identifiers: top_identifiers
         )
@@ -90,11 +89,6 @@ module Page
                 .group_by { |i| i[:scheme] }
                 .sort_by { |s, ids| [-ids.size, s] }
                 .map { |s, _ids| s }
-    end
-
-    def image_proxy_url(id)
-      'https://mysociety.github.io/politician-image-proxy' \
-      "/#{country.slug}/#{house.slug}/#{id}/140x140.jpeg"
     end
 
     # Caches for faster lookup
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -60,7 +60,7 @@ module Page
         PersonCard.new(
           person:          person,
           proxy_image:     image_proxy_url(person.id),
-          memberships:     person_memberships(person),
+          term:            term,
           top_identifiers: top_identifiers
         )
       end
@@ -92,20 +92,12 @@ module Page
                 .map { |s, _ids| s }
     end
 
-    def person_memberships(person)
-      membership_lookup[person.id]
-    end
-
     def image_proxy_url(id)
       'https://mysociety.github.io/politician-image-proxy' \
       "/#{country.slug}/#{house.slug}/#{id}/140x140.jpeg"
     end
 
     # Caches for faster lookup
-    def membership_lookup
-      @membership_lookup ||= current_term_memberships.group_by(&:person_id)
-    end
-
     def area_lookup
       @area_lookup ||= popolo.areas.group_by(&:id)
     end
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -60,7 +60,6 @@ module Page
         PersonCard.new(
           person:          person,
           term:            term,
-          top_identifiers: top_identifiers
         )
       end
     end
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -56,7 +56,7 @@ module Page
     end
 
     def people
-      @people ||= people_for_current_term.sort_by { |e| [e.sort_name, e.name] }.map do |person|
+      @people ||= term.people.sort_by { |e| [e.sort_name, e.name] }.map do |person|
         PersonCard.new(
           person:          person,
           term:            term,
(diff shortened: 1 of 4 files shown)


lib/person_cards.rb
```diff
@@ -39,6 +41,10 @@ def identifiers
     Section::Identifiers.new(person, top_identifiers: top_identifiers).data
   end
 
+  def memberships
+    @person.memberships.select { |m| m.term.id == @term.id }
```
tmtmtmtm: Rather than rolling your own `select` here, it's probably better to go through the public search interface: i.e. `@person.memberships.where(legislative_period_id: term.id)`

lib/page/term_table.rb
```diff
@@ -3,6 +3,35 @@
 require_relative '../everypolitician_extensions'
 require_relative '../person_cards'
 
+# Reopen the LegislativePeriod class to add helper methods; we could
+# consider adding some of these to the class definition in
+# everypolitician-ruby instead.
```
tmtmtmtm: Rather than re-opening here, I think it would be better to add these into `lib/everypolitician_extensions.rb`, especially as we're already extending `LegislativePeriod` there.

mhl: Thanks, I hadn't spotted that (despite actually using the `memberships` method defined there!)

lib/person_cards.rb
```diff
@@ -1,14 +1,16 @@
...
-  # TODO: pass fewer arguments!
-  def initialize(person:, proxy_image:, memberships:, top_identifiers:)
+  def initialize(person:, term:)
     @person          = person
-    @memberships     = memberships # should be able to get from Person
-    @proxy_image     = proxy_image # get from Legislature
-    @top_identifiers = top_identifiers # get from Legislature
+    @term            = term
+    @legislature     = term.legislature
+    @top_identifiers = term.top_identifiers
```
tmtmtmtm: NB: this is now better documented: https://github.com/everypolitician/viewer-sinatra/blob/master/CONTRIBUTING.md

mhl:
Thanks for all those helpful comments, @tmtmtmtm - it was too confusing to amend these via fixups so I've pushed a new version of the branch.

I've tried diffing this generated version with the other and there are some changes:
1. Sometimes the order of two people with an identical `sort_name` is reversed (e.g. Octavio Jara Wolff and Sergio Jara Catalán)
2. Sometime the top identifiers are in a different order. In the examples I've looked at this is similarly because there there the same number of those identifiers.

These are both due to the following returning people in different orders:

```
# On master:
popolo.persons.select { |p| Set.new(term.memberships.map(&:person_id)).include?(p.id) }
```

```
# On simplify-personcard
term.memberships.map(&:person).uniq(&:id)
```

The top identifiers and persons are futher sorted before use, so this only has an effect where the sort key is the same for two different items. The differences in output are pretty inconsequential, I think, but to potentially reduce the time spent in looking into similar problems in the future, I'd be tempted to add:

```diff
diff --git a/lib/everypolitician_extensions.rb b/lib/everypolitician_extensions.rb
index 6928a06..c8ef4d7 100644
--- a/lib/everypolitician_extensions.rb
+++ b/lib/everypolitician_extensions.rb
@@ -49,7 +49,7 @@ module EveryPolitician
                            .flatten
                            .reject { |i| i[:scheme] == 'everypolitician_legacy' }
                            .group_by { |i| i[:scheme] }
-                           .sort_by { |_s, ids| -ids.size }
+                           .sort_by { |s, ids| [-ids.size, s] }
                            .map { |s, _ids| s }
     end
   end
diff --git a/lib/page/term_table.rb b/lib/page/term_table.rb
index 0e9e0b4..4e0c28d 100644
--- a/lib/page/term_table.rb
+++ b/lib/page/term_table.rb
@@ -56,7 +56,7 @@ module Page
     end
 
     def people
-      @people ||= term.people.sort_by(&:sort_name).map do |person|
+      @people ||= term.people.sort_by{|e| [e.sort_name, e.name]}.map do |person|
         PersonCard.new(
           person: person,
           term:   term
```

What do you think?

tmtmtmtm:
@mhl yes, I think stabilising the sort order is a good idea.

lib/person_cards.rb
```diff
@@ -23,6 +23,14 @@ def image
...
   end
 
+  def legislature
+    term.legislature
+  end
+
+  def top_identifiers
+    term.top_identifiers
+  end
+
```
tmtmtmtm: Sorry for not catching this earlier, but I'm not convinced that these need to be public, do they?

mhl: Good point - thanks. I've force-pushed a version that changes those two methods to be private.

mhl: Oh, except probably `memberships` should be private too, grr

mhl: Ah, no - it's used by the term_table.erb view.

@tmtmtmtm I think / hope this is OK now!

Issue #15606, opened 2017-02-17
Add method for accessing legislature position data


chrismytton:
# Problem

As part of https://github.com/everypolitician/viewer-sinatra/issues/24 we'll need to get access to the `unstable/positions.csv` file in the legislature's directory in ep-data, at least until we move that data into the Popolo.

# Proposed Solution

The [code I wrote for the prototype](https://github.com/everypolitician/viewer-sinatra/blob/01201e4ca12e5a28d3730d9cc8f3f0a6d18236d4/lib/page/term_table.rb#L83-L85) does the job, so possibly just a case of pulling this into `lib/everypolitician_extensions.rb` and tidying it up a bit.

We should also have a `Struct` that represents each individual position object, so we don't have to use `[]` to access properties of a position.

# Acceptance Criteria

I should be able to access the positions for a legislature in a `Page::TermTable` instance by calling `house.positions`.

# Related Issues

- Executive memberships https://github.com/everypolitician/viewer-sinatra/issues/24
- Work out how to display executive positions when known  https://github.com/everypolitician/viewer-sinatra/issues/13297

chrismytton:
I've done a prototype of this in https://github.com/everypolitician/viewer-sinatra/pull/15608 but it's currently blocked because the test fixtures have an old version of the position CSVs which is missing the `type` column. I've opened #15607 to track that.

Issue #24, opened 2015-04-21
Executive memberships


tmtmtmtm:
Display Executive-level memberships per Person, and per-Government, including pivoting to a view of who else has held that same position over time.

tmtmtmtm:
Depends on https://github.com/everypolitician/everypolitician/issues/178

chrismytton:
Related to https://github.com/everypolitician/everypolitician/issues/178

chrismytton:
Have done a prototype of this in https://github.com/everypolitician/viewer-sinatra/pull/13600.

Things that came out of the prototype:

- [ ] Need to make accessing the `unstable/positions/csv` file a little less involve. For now this complexity can probably just be hidden in a class, then it can be moved to ep-ruby at a later point.
- [ ] Make sure we filter the list of positions to only those that have `type = cabinet`  in the CSV
- [ ] Need to filter positions to show only those for the current term
- [ ] Probably best to add position data to the person class, rather than passing another constructor argument to `PersonCard`

The prototype was useful as a guiding light, but I think for the next stage we probably want to throw that away and do a fresh version using TDD.

chrismytton:
# TODO

- [ ] Add classes to the `lib/everypolitician_extensions.rb` file for interacting with position data. Prototype of this in https://github.com/everypolitician/viewer-sinatra/pull/15608.
  - [ ] `Everypolitician::Position` which represents an individual position held
  - [ ] `Everypolitician::Positions` for filtering and interacting with positions data
    - [ ] Should be able to filter by `type` (we just want to show the `cabinet` positions initially)
    - [ ] Should be able to filter by term, i.e. "give me all the positions between these two dates"
- [ ] Add `Everypolitician::UnstablePositionCSV` class for getting the position data out of the `unstable/positions.csv` file. https://github.com/everypolitician/viewer-sinatra/issues/15606
  - This can then be replaced with another class when we decide where the position data will live.
- [ ] Allow `Everypolitician::Popolo::Person` instances to access the position data for the person

PR #15609, opened 2017-02-21
Add README for unstable position data interface

chrismytton:
# What does this do?

This adds a section to the README with details on the expected interface for using the unstable position data.

# Why was this needed?

We want to add an interface for interacting with unstable position data to viewer-sinatra. This interface is an extension to the `EveryPolitician::Popolo::Person` class, which operates on the
`unstable/positions.csv` file in a legislature's directory in everypolitician-data.

# Relevant Issue(s)

Part of https://github.com/everypolitician/viewer-sinatra/issues/15606

- High level issue: Executive memberships https://github.com/everypolitician/viewer-sinatra/issues/24

# Implementation notes

I was originally going to add the `positions` method to the `Legislature`, but that doesn't really make sense, as it's people that hold positions, so that's where I've ended up putting the method.

In the (very near) future we'll be adding methods for filtering positions to just those that fall between two dates, so that we can show positions for a particular term. But for now this interface should give us enough to get off the ground.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,38 @@ bundle exec rake test:all
 bundle exec rake
 ```
 
+## Internal documentation
+
+### Interacting with unstable data
+
+We use data from the `unstable/positions.csv` file in a legislature's
+directory to display cabinet memberships on the website. Because this data
+is unstable there currently aren't methods in [the `everypolitician`
+gem](https://github.com/everypolitician/everypolitician-ruby) for
+interacting with it, so viewer-sinatra has some of its own classes for
+dealing with it. These classes live in the
+`lib/everypolitician_extensions.rb` file and are documented below.
+
+#### Get all cabinet memberships for a person
+
+To print all of the known cabinet memberships for a person, you can do the
+following:
+
+```ruby
+house_of_commons = Everypolitician::Index.new.country('UK').legislature('Commons')
+person = house_of_commons.popolo.persons.find_by(name: 'Gordon Brown')
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,6 +110,31 @@ bundle exec rake test:all
 bundle exec rake
 ```
 
+## Internal documentation
+
+### Interacting with unstable position data
+
+We use data from the `unstable/positions.csv` file in a legislature's directory to display cabinet positions on the website. Because this data is unstable there currently aren't methods in [the `everypolitician` gem](https://github.com/everypolitician/everypolitician-ruby) for interacting with it, so viewer-sinatra has some of its own classes for dealing with it. These classes live in the `lib/everypolitician_extensions.rb` file and are documented below.
```
tmtmtmtm: not specifically an issue here, but just for explicit clarity, it's not only because the file lives in `unstable/` that we can't interact with it from the gem — it's that the way we store this data *will* change. The `positions.csv` is just a very quick hack to let us get up and running — we still need to work out how exactly we're going to store this sort of information, and the answer may not even be within the legislature at all.

chrismytton:
> In principle this looks good, though it would be better to have an example that actually exists in the current positions.csv data, which was migrated to the new approach (and now only holds Cabinet data) in everypolitician/everypolitician-data#26264

Ah yep, I've updated the example to use the latest data.

As discussed in Slack, I've also changed this to talk about memberships instead of positions and to make this specifically about adding a `cabinet_positions` method.

PR #15610, opened 2017-02-21
Reduce duplication of fixture SHAs in tests

### LINKED ISSUE #15607: Make it easier to update test fixtures
# Problem

At the moment updating the test fixtures involves lots of bitty steps that are easy to get wrong. This means that the tests run against quite an old version of everypolitician-data. It would be good to make this easier so we can test the app against newer versions of the data.

# Proposed Solution

Add a `rake` task which updates `t/fixtures/countries.json` (which is currently called `d8a4682f-countries.json`) and the files in `t/fixtures/everypolitician-data` to the latest version on the everypolitician-data `master` branch.

This will involve going through the output from `git grep stub_ t/` and removing any hard-coded SHAs and moving those into a single central data file. This way the SHAs are tracked under source control but don't need to be updated in the individual tests each time.

# Acceptance Criteria

Updating the test fixtures to the latest version should only require one command.

# Related Issues

This is needed in order to do https://github.com/everypolitician/viewer-sinatra/issues/15606 because the current version of the fixtures don't have the `type` column in the `unstable/positions.csv`, so it's awkward to add a test for the `Position#type` method.

chrismytton:
## What does this do?

This reduces the number of places that we're storing the SHA of the fixtures in the tests.

## Why was this needed?

Previously we were storing the SHA in the following places:

- In the VCR cassettes
- In the `countries.json` fixture
- In the directory names of the popolo fixtures
- In the tests themselves, with code like `stub_popolo('faa324', 'UK/Commons')`

This meant that updating the fixtures to a newer version (which we want to do as part of adding cabinet memberships #24, and probably again when we finalise where that data is going to live), was quite a complicated and hard to script process.

So this change removes the SHA from the fixture directory name and the tests themselves, which should be enough to make updating the fixtures relatively painless.

## Relevant Issue(s)

Fixes #15607 

Unblocks #15606 

## Notes to Reviewer

As part of this change I've also switched some places that were still using `stub_everypolitician_data_request` to use the newer `stub_popolo` method. I've done this in a separate commit so that can be split off into a separate pull request if preferred.


chrismytton:
Have popped another commit on the end of this to fix the rake which downloads fixtures. So I think this is _actually_ ready for review now.

chrismytton:
This is needed because the tests are currently using the `index_at_known_sha` method to access countries and legislatures. This method uses the local `countries.json` fixture to determine which SHA to use for a given legislature. So even though we can have different legislatures at different SHAs, most of the tests go through `countries.json`, so will use that SHA for the legislature.

tmtmtmtm:
@chrismytton I suspect it's probably best if you could resolve all the conflicts before this gets reviewed…

PR #15612, opened 2017-02-23
Update README docs for unstable cabinet memberships

chrismytton:
# What does this do?

Updates the README with a new interface for accessing the cabinet positions for a person.

# Why was this needed?

Trying to handle unstable cabinet memberships directly in the Popolo classes wasn't working because the Popolo classes (understandably) don't know anything about the world outside Popolo.

So to fix this we're going to move the logic for looking up cabinet memberships into the `TermTable` and `PersonCard` classes for now. Then once we've found a stable place for the cabinet memberships to live we can then move more of the logic into the Popolo classes or wherever makes sense.

# Relevant Issue(s)

Part of #15606 

The original pull request where I added the README for this was https://github.com/everypolitician/viewer-sinatra/pull/15609

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,7 @@ directory to display cabinet memberships on the website. Because this data
 is unstable there currently aren't methods in [the `everypolitician`
 gem](https://github.com/everypolitician/everypolitician-ruby) for
 interacting with it, so viewer-sinatra has some of its own classes for
-dealing with it. These classes live in the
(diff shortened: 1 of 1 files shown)


chrismytton:
The README doesn't seem like the right place for these docs, so closing this pull request in favor of https://github.com/everypolitician/viewer-sinatra/pull/15616.

PR #15615, opened 2017-02-24
Add cabinet memberships to term table

### LINKED ISSUE #24: Executive memberships

chrismytton:
# What does this do?

Outputs cabinet memberships on term table page.

# Why was this needed?

We want to output cabinet memberships a member has held in order to finish #24.

# Relevant Issue(s)

Fixes #24 

# Implementation notes

This adds some methods to the `lib/everypolitician_extensions.rb` file which augments the everypolitician-ruby and everypolitician-popolo gems. Currently the logic for connecting a person to a cabinet membership is in the `PersonCard` class, as it doesn't really make sense trying to shoehorn it into the Popolo classes.

# Screenshots

cabinet-memberships

# Notes to Reviewer

- I'm not very happy with the tests in this pull request. I'm open to suggestions on how these can be improved.
- The tests are still using quite an old version of `countries.json`, so it's testing against an older format of CSV
- This only filters out positions with no `start_date`, it doesn't do anything about the `type` column. I don't _think_ this is  an issue because `positions.csv` now only holds cabinet data (as per https://github.com/everypolitician/viewer-sinatra/pull/15609#pullrequestreview-22923927).
- It's quite possible I'm trying to change too much at once in this pull request. Happy to try splitting it up if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,39 +110,6 @@ bundle exec rake test:all
 bundle exec rake
 ```
 
-## Internal documentation
-
-### Interacting with unstable data
-
-We use data from the `unstable/positions.csv` file in a legislature's
-directory to display cabinet memberships on the website. Because this data
-is unstable there currently aren't methods in [the `everypolitician`
-gem](https://github.com/everypolitician/everypolitician-ruby) for
-interacting with it, so viewer-sinatra has some of its own classes for
-dealing with it. These classes live in the
-`lib/everypolitician_extensions.rb` file and are documented below.
-
-#### Get all cabinet memberships for a person
-
-To print all of the known cabinet memberships for a person, you can do the
-following:
-
-```ruby
-house_of_commons = Everypolitician::Index.new.country('UK').legislature('Commons')
-person = house_of_commons.popolo.persons.find_by(name: 'Gordon Brown')
-
-person.cabinet_memberships.each do |membership|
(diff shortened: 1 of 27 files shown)


tmtmtmtm:
> I'm not very happy with the tests in this pull request. I'm open to suggestions on how these can be improved.

One of the advantages of testing against fixed data is that you can write concrete tests, rather than abstract ones — i.e. pick a few people with long, complex membership history, and make sure they have the correct memberships in each of a few terms. 

> This only filters out positions with no start_date, it doesn't do anything about the type column. I don't think this is an issue because positions.csv now only holds cabinet data (as per #15609 (review)).

I think that's a misreading of that comment. That's saying that the *UK* `positions.csv` only holds cabinet data — in many other cases (including ones that you've [written tests against](https://github.com/everypolitician/viewer-sinatra/pull/15615/commits/45317e8e981cdc0e1538ec64a521cce6f5135d00)), there are other positions too.

> It's quite possible I'm trying to change too much at once in this pull request. Happy to try splitting it up if needed.

Yep, you definitely are :)

Things like changing how the stubbing in tests works could go through separately in advance, as could the docs, and probably the library extensions.

PR #15616, opened 2017-02-28
Cabinet memberships documentation

chrismytton:
# What does this do?

Updates the documentation for the cabinet memberships methods that was added in https://github.com/everypolitician/viewer-sinatra/pull/15609 and moves it out of the README and into `lib/person_cards.rb`.

# Why was this needed?

Originally we thought this code might be useful to outside people, as it was augmenting the everypolitician-ruby gem. Now though, we've decided that it's best for the functionality to live in some viewer-sinatra specific classes for now, until we have a better idea of how it all fits together at least. So this moves the docs to a more logical place.

# Relevant Issue(s)

Extracted from https://github.com/everypolitician/viewer-sinatra/pull/15615

Follows on from https://github.com/everypolitician/viewer-sinatra/pull/15609

Replaces https://github.com/everypolitician/viewer-sinatra/pull/15612

Part of #24 

# Implementation notes

This uses the YARD `@example` documentation tag, which render the example code out in the docs, see screenshot below.

# Screenshots

yard-person-card

# Notes to Reviewer

This is just the documentation for now, the implementation will come as a separate pull request.

PR #15619, opened 2017-02-28
Rename PersonCard#memberships to #legislative_memberships

chrismytton:
# What does this do?

This renames the `PersonCard#memberships` to `PersonCard#legislative_memberships`.

# Why was this needed?

We want to add another method to the `PersonCard` class which returns cabinet memberships as part of #24. Therefore it makes sense to rename this method to legislative_memberships to make it clear that it's only returning a subset of the memberships.

# Relevant Issue(s)

Extracted from https://github.com/everypolitician/viewer-sinatra/pull/15616

Part of #24 

# Implementation notes

~~Since this is an internal class I've decided to just rename the method, rather than bother keeping an alias to the old method as I originally did in https://github.com/everypolitician/viewer-sinatra/pull/15616.~~ 

I realised that just renaming the method would mean changing code and tests in one pull request, which is a Bad Thing. Instead, I've left the original `memberships` method in place as an alias to the new `legislative_memberships` method. The `memberships` method should be removed in a future pull request.

### DIFF
--- a/lib/person_cards.rb
+++ b/lib/person_cards.rb
@@ -58,11 +58,14 @@ class PersonCard
     Section::Identifiers.new(person, top_identifiers: top_identifiers).data
   end
 
-  # List of memberships this person held in this term
+  # List of legislative memberships this person held in this term
   # @return [Array<EveryPolitician::Popolo::Membership>]
-  def memberships
+  def legislative_memberships
     person.memberships.where(legislative_period_id: term.id)
   end
+
+  # TODO: Remove this alias and the associated tests.
+  alias memberships legislative_memberships
 
   private
 
--- a/t/page/term_table.rb
+++ b/t/page/term_table.rb
@@ -95,6 +95,14 @@ describe 'TermTable' do
         af.memberships.first.area.name.must_equal 'Wahlkreis: 3B – Waldviertel'
       end
 
+      it 'has a single legislative membership' do
+        af.legislative_memberships.count.must_equal 1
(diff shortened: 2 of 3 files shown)


lib/person_cards.rb
```diff
@@ -58,12 +58,15 @@ def identifiers
...
-  # List of memberships this person held in this term
+  # List of legislative memberships this person held in this term
   # @return [Array<EveryPolitician::Popolo::Membership>]
-  def memberships
+  def legislative_memberships
     person.memberships.where(legislative_period_id: term.id)
   end
 
+  # TODO: Remove this alias and the associated tests.
+  alias memberships legislative_memberships
```
tmtmtmtm: The "implementation notes" section of the PR description claims "Since this is an internal class I've decided to just rename the method, rather than bother keeping an alias to the old method", but the code disagrees…

chrismytton: Sigh, yes, I forgot to update that after I changed the PR 😞  have changed the "Implementation notes" section to match reality now.

chrismytton:
After discussing with @tmtmtmtm in Slack we've decided that this isn't a refactoring, since it's adding a new method. I've opened https://github.com/everypolitician/viewer-sinatra/pull/15620 which is the same code as this pull request, but without `refactor` in the branch name so the `check-fail` script will run.

Closing this in favor of https://github.com/everypolitician/viewer-sinatra/pull/15620

PR #15621, opened 2017-03-01
Add cabinet methods to Everypolitician extensions

chrismytton:
# What does this do?

Adds `Legislature#cabinet_memberships` and `LegislativePeriod#cabinet_memberships` method to the `lib/everypolitician_extensions.rb` file.

# Why was this needed?

These methods are needed as part of #24 so that we can access the cabinet memberships from the `PersonCard` instance and display them on the term table page.

# Relevant Issue(s)

Part of #24 

Extracted from https://github.com/everypolitician/viewer-sinatra/pull/15615


chrismytton:
I've removed the `Legislature#cabinet_positions` method and pushed the logic into the `LegislativePeriod` extension.

lib/everypolitician_extensions.rb
```diff
@@ -52,6 +58,32 @@ def top_identifiers
                            .sort_by { |s, ids| [-ids.size, s] }
                            .map { |s, _ids| s }
     end
+
+    private
+
+    def current_cabinet_memberships
```
tmtmtmtm: I'm not entirely sure what this method is meant to be for, and the implementation is a little hard to follow, but I _think_ the "current" is a nismomer.

lib/everypolitician_extensions.rb
```diff
@@ -52,6 +58,32 @@ def top_identifiers
...
+      all_cabinet_memberships.select do |mem|
+        !mem.start_date.nil? && (mem.end_date.to_s.empty? || mem.end_date >= start_date.to_s)
```
tmtmtmtm: I've been peering at this for quite a while now, and it's making my brain hurt. 

I can't quite tell whether `!mem.start_date.nil?` is meant to be doing something different from a plain `mem.start_date`, nor what the `&&` condition is actually checking for.

I suspect this might be clearer as a `reject`, or possibly even more than one `reject`, rather than a `select`, but it'd almost certainly be even clearer still moving these to named methods on  `CabinetMembership`.

Issue #15611, opened 2017-02-22
show example code snippets, customised for each legislature


davewhiteland:
# Problem

We want coders to know it's easy to get this data using the library (for now, let's just say Ruby). 

(If this looks cool, adding a "change language" bit of JS magic can be another issue)

# Proposed Solution

Add snippet of example code to legislature pages which gets the data; significantly includes the country and house names in the code (since obviously we know them for each page we're generating... actually we're generating them using the library itself heh).

Snippet explained with something like

```Ruby
      # Use this data in your own Ruby app like this:
      require 'everypolitician'
      country = Everypolitician::Index.new.country('Australia')
      house = australia.legislature('Senate')
   
     #   See: https://github.com/everypolitician/everypolitician-ruby
```

# Acceptance Criteria

* snippet of cut-and-pasteable code is shown together with brief explanation as shown
* snippet runs correctly where e.g. gem installed

# Not Required

Other libraries/languages also supported: ignore for now

# Prerequisites

none

# Related Issues

none

# Due Date

none, so ASAP

# Design?

yes

# Text?

A little bit

# Bloggable?

Yes

kudos to @struan for the suggestion

PR #15624, opened 2017-03-02
Show Ruby code example on every term-table page

davewhiteland:
# What does this do?

Adds an example snippet of Ruby (using the everypolitician gem) for accessing the data that's displayed on a term-table page.

# Why was this needed?

Enhancement: if a potential developer is looking at that page on the EveryPolitician website, on the basis that a simple example is a dev's favourite favourite way of learning, then this communicates how easy it is to use the data in their own project.

# Relevant Issue(s)

#15611 

# Implementation notes

None: no design effort made here (e.g. syntax colouring, any dynamic hide/show etc)... this just uses an existing section style, with a new class (`page-section--code`) that has no explicit CSS associated with it — in fact this class is only used in order to isolate the code example the tests.

# Screenshots

screen shot 2017-03-02 at 16 08 52

# Notes to Reviewer

The output from running the snippet of code is non-specific: just `name (party): gender` which is enough to convey that the program has access to the data, iterating over the memberships. It's _almost_ the same as the one we used on the EveryPolitician leaflet—the difference is _this_ code uses the specific term ID, and prints the gender (not the twitter handle) simply because without needing further checks gender is more likely to be populated with data.

Also: it's not obvious what the right targets should be for the links to the libraries: check that they are what you expect

* Ruby https://github.com/everypolitician/viewer-sinatra/blob/7bf8de9df72ae271a2dbd80fb27e1f96bd8571fd/views/term_table.erb#L216-L217
* Python https://github.com/everypolitician/viewer-sinatra/blob/7bf8de9df72ae271a2dbd80fb27e1f96bd8571fd/views/term_table.erb#L230
* PHP https://github.com/everypolitician/viewer-sinatra/blob/7bf8de9df72ae271a2dbd80fb27e1f96bd8571fd/views/term_table.erb#L231

Finally here's the example cut-and-pasted from a local run. _This_ code (that is, the example being created in the HTML) is not tested automatically in Viewer-Sinatra's tests; this snippet works for me locally in my `everypolitician-data` bundled environment.

```Ruby
require 'everypolitician'
country = Everypolitician::Index.new.country('Japan')
house = country.legislature('House-of-Representatives')
house.popolo.terms.find_by(id: 'term/46').memberships.each do |m|
  puts "#{m.person.name} (#{m.party.name}): #{m.person.gender}"
end
```

sample output (truncated):

```
KAWAMURA Takeo (LDP): male
MORIYAMA Masahito (LDP): male
NAKAGAWA Yuko (LDP): female
NAGAOKA Keiko (LDP): female
HASHIMOTO Hidenori (LDP): male
...
```

# Notes to Merger

Only consequence is an extra div containing the example code at the bottom of every term-table page.

### DIFF
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -221,8 +221,8 @@
 require 'everypolitician'
 country = Everypolitician::Index.new.country('<%= @page.country.slug %>')
 house = country.legislature('<%= @page.house.slug %>')
-house.popolo.terms.find_by(id: '<%= @page.current_term.id %>').memberships.each do |m|
-  puts "#{m.person.name} (#{m.party.name}): #{m.person.gender}"
+house.popolo.terms.find_by(id: '<%= @page.current_term.id %>').memberships.each do |membership|
+  puts "#{membership.person.name} (#{membership.party.name}): #{membership.person.gender}"
 end
           </pre>
           <p>
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -223,8 +223,7 @@ country = Everypolitician::Index.new.country('<%= @page.country.slug %>')
 house = country.legislature('<%= @page.house.slug %>')
 house.popolo.terms.find_by(id: '<%= @page.current_term.id %>').memberships.each do |membership|
   puts "#{membership.person.name} (#{membership.party.name}): #{membership.person.gender}"
-end
-          </pre>
+end</pre>
           <p>
             Libraries also available in
             <a href="https://github.com/everypolitician/everypolitician-python">Python</a> and 
--- a/views/term_table.erb
+++ b/views/term_table.erb
@@ -222,7 +222,7 @@ require 'everypolitician'
 country = Everypolitician::Index.new.country('<%= @page.country.slug %>')
 house = country.legislature('<%= @page.house.slug %>')
 house.popolo.terms.find_by(id: '<%= @page.current_term.id %>').memberships.each do |m|
-  puts "#{m.person.name} (#{m.party.name}): @#{m.person.twitter || '?'}"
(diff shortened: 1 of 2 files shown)


views/term_table.erb
```diff
@@ -209,6 +209,29 @@
...
+            like this:
+          </p>
+          <pre>
+require 'everypolitician'
+country = Everypolitician::Index.new.country('<%= @page.country.slug %>')
+house = country.legislature('<%= @page.house.slug %>')
+house.popolo.terms.find_by(id: '<%= @page.current_term.id %>').memberships.each do |m|
+  puts "#{m.person.name} (#{m.party.name}): #{m.person.gender}"
+end
+          </pre>
```
chrismytton: It's probably best to put this `</pre>` tag _immediately_ after the `end` on the line above. This should prevent the extra line break that's currently appearing when its rendered.

views/term_table.erb
```diff
@@ -209,6 +209,29 @@
...
+            Ruby programmer? It’s easy to use this data in your own app! Use the
+            <a href="https://medium.com/@everypolitician/introducing-the-everypolitician-gem-ab39db5b7e49">gem</a>
+            <!-- https://github.com/everypolitician/everypolitician-ruby -->
+            like this:
+          </p>
+          <pre>
+require 'everypolitician'
+country = Everypolitician::Index.new.country('<%= @page.country.slug %>')
+house = country.legislature('<%= @page.house.slug %>')
+house.popolo.terms.find_by(id: '<%= @page.current_term.id %>').memberships.each do |m|
```
chrismytton: I think it's worth giving the block parameter a better name for this example. Perhaps `membership` rather than `m`?

Issue #15625, opened 2017-03-06
Run all tests in `rake test` task


chrismytton:
# Problem

Running `rake test` doesn't run all of the tests. This is surprising, because our other Ruby apps/libs run all tests when `rake test` is run. This is a problem because the `ensure-regression-tests` script expects all tests to be run when `rake test` is run.

# Proposed Solution

Rename the `test` task to be `test:page` and then rename the `test:all` task to be `test`. We'll also probably need to update `.travis.yml`.

# Acceptance Criteria

Running `rake test` should run _all_ tests in the `t` directory.

PR #15626, opened 2017-03-06
Rename rake test:all to test

chrismytton:
# What does this do?

Renames the `test:all` rake task to be called `test`.

# Why was this needed?

The `ensure-regression-tests` script that gets run on Travis assumes that all of your tests will run when `rake test` is executed. This wasn't the case for this repo, which means that the script was failing on Travis, even thought there were regression tests.

# Relevant Issue(s)

Fixes https://github.com/everypolitician/viewer-sinatra/issues/15625

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,7 @@ require 'rubocop/rake_task'
 require 'reek/rake/task'
 
 Rake::TestTask.new do |t|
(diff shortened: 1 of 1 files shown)


chrismytton:
Forgot to put "refactor" in the branch name 😞

PR #15633, opened 2017-12-13
Update gems

tmtmtmtm:
Update all gems to latest versions in preparation for new work.


henare:
The build failure [seen here](https://travis-ci.org/everypolitician/viewer-sinatra/builds/316083803#L796) should be fixed by 069ff9a in #15634. Once that PR is merged this could be revisited.

tmtmtmtm:
I've rebased this over the new post-15634 `master` so hopefully things are a bit happier now.

tmtmtmtm:
Ah, nope, various other things have changed underneath us, so we'll need to do this piecemeal. Looks like rubocop is the main failure now, but it's in the config inherited from `everypolitician-data` so we'll need to fix it over there.

tmtmtmtm:
I updated the `rubocop` base in https://github.com/everypolitician/everypolitician-data/pull/68318 and then ran `rubocop -a` here.

tmtmtmtm:
Woohoo!

PR #15634, opened 2017-12-14
Fix Travis build

henare:
# What does this do?

Gets Travis building `master` successfully again.

# Why was this needed?

`master` cannot currently be built on Travis. We need successful builds to be able to continue to make changes.

# Relevant Issue(s)

N/A

# Implementation notes

Two problems were preventing the build from working.

Travis seems to now create a `~/bin` folder so the [`scripts/install_tidy.sh` script was failing](https://travis-ci.org/everypolitician/viewer-sinatra/builds/316179211#L796). 1500da4 fixes this by not failing if the directory already exists.

Two gems had new security advisories so [`rake bundle:audit` was failing](https://travis-ci.org/everypolitician/viewer-sinatra/builds/316180819#L960-L976). 87884b8 and 069ff9a fix this by updating the two gems.

# Screenshots

N/A

# Notes to Reviewer

N/A

# Notes to Merger

N/A

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,11 +47,11 @@ GEM
     iso_country_codes (0.7.1)
     json (1.8.2)
     method_source (0.8.2)
-    mini_portile2 (2.2.0)
+    mini_portile2 (2.3.0)
     minitest (5.9.0)
     multipart-post (2.0.0)
-    nokogiri (1.8.0)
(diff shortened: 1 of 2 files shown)


henare:
>(NB: changes like 1500da4 would generally be better with a bit more explanation in the commit message)

@tmtmtmtm Thanks for reviewing and merging, and thanks for pulling me up on this :+1:

PR #15636, opened 2017-12-15
Install sinatra reloader

henare:
[Reopening of #15635 to include `refactor` in branch name to make `ensure-regression-tests` happy.]

# What does this do?

This reloads files that have been changed in development. 

# Why was this needed?

It's useful for us when app.rb has been changed because prior to this you needed to
manually restart the server if you changed this file.

# Relevant Issue(s)

N/A

# Implementation notes

N/A

# Screenshots

N/A

# Notes to Reviewer

Includes `refactor` in branch name so `ensure-regression-tests` passes. Since this is only introducing functionality in development we don't expect there to be new tests that fail against master.

# Notes to Merger

N/A

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,10 @@ gem 'sass'
 gem 'sinatra'
 gem 'yajl-ruby', require: 'yajl'
 
+group :development do
+  gem 'sinatra-contrib'
+end
+
 group :test do
   gem 'bundler-audit'
   gem 'html_validation'
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ Finally, follow the instructions displayed by the virtual machine.
 
 Sass files are currently generated on each pageload, just like any other template. You do not need to manually compile the Sass.
 
+In development, changes to `app.rb` and Ruby files in the `lib` directory are automatically reloaded by  [`Sinatra::Reloader`](http://sinatrarb.com/contrib/reloader.html). You do not need to restart the webserver process.
+
 ### Run without Vagrant
 
 You can also run this project locally without using Vagrant. First [install Foreman globally](https://github.com/ddollar/foreman#installation):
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -48,8 +48,6 @@ First install Vagrant following [the steps in the official documentation](http:/
 
 Finally, follow the instructions displayed by the virtual machine.
 
-**Note:** Changes to `app.rb` will require you to restart the webserver process.
```
tmtmtmtm: Rather than simply removing this, I think it might be better to state that new reality explicitly, especially as the new behaviour doesn't seem to be obvious, and people's won't necessarily expect that this will happen.

Gemfile
```diff
@@ -22,6 +22,10 @@ gem 'sass'
 gem 'sinatra'
 gem 'yajl-ruby', require: 'yajl'
 
+group :development do
+  gem 'sinatra-contrib'
```
tmtmtmtm: According to https://github.com/sinatra/sinatra-contrib this has been merged into sinatra core now. My assumption is that this version still works as it's also still available as the old standalone gem, but it's not clear that any updates etc are backported to there. (Though I haven't looked closely at this, and might be misunderstanding entirely)

henare: Well spotted but it's just that the source code repos have been merged, not the gems.

https://rubygems.org/gems/sinatra-contrib/versions is still being released standalone and you can also see that the `sinatra` gem still [doesn't include the contrib files](https://github.com/sinatra/sinatra/blob/master/sinatra.gemspec#L10-L21).

README.md
henare: Good call @tmtmtmtm :+1:

I've added this back in 15a615b and moved this and the note about Sass to a more appropriate section in 7c4ced4 (tidying as I go).

PR #15637, opened 2017-12-20
Display seat count from Wikidata on legislature wikidata subpages

henare:
# What does this do?

Shows the legislature seat count from Wikidata on the not-currently-scraped `/:country/:house/wikidata` pages.

# Why was this needed?

This was suggested as the [tiny first step to building the Democratic Commons' Legislature Browser](https://github.com/everypolitician/democratic-commons-tasks/issues/36#issuecomment-351529404).

# Relevant Issue(s)

https://github.com/everypolitician/democratic-commons-tasks/issues/36#issuecomment-351529404

# Implementation notes

[Capybara](https://github.com/teamcapybara/capybara) and [capybara-webkit](https://github.com/thoughtbot/capybara-webkit) have been installed to allow testing the page with JavaScript enabled. It's also envisioned that we could use this to snapshot the page when we include it in the static site.

# Screenshots
What it looks like on the page:

screenshot from 2017-12-20 18-04-12

screenshot from 2017-12-21 15-06-06

Hokey loading thing while it's fetching the data from the Wikidata API:

screenshot from 2017-12-20 18-13-33

# Notes to Reviewer

It would be particularly useful to know if this approach is looking like it works, especially the JavaScript.

It would also be useful to get some thoughts about the user experience, e.g. the loading... placeholder is particularly hokey - any suggestions for different ways to do this would be helpful.

# Notes to Merger

N/A


mhl:
Hi @henare - I'm afraid I don't have time to look at the detail of this right now, but I hope it's OK to make a few general comments:

There's a lot of commits here that really represent work-in-progress which is subsumed by later commits. These intermediate commits are _sometimes_ interesting, but as a reviewer, I'd almost always prefer to have a small pull request where any work in progress has been squashed out. Looking at the overall diff that this pull request introduces it's actually ended up being quite a reasonably sized change, so I think you can easily rewrite the history of this branch to be a much smaller number of commits. It'll also make the git history much easier to understand in future, and helps reviewers who go commit-by-commit (as I do).

I'm afraid this actually isn't articulated properly in https://mysociety.github.io/coding-standards.html at the  moment, but we should add something about that there!

henare:
Thanks @mhl, that's helpful :+1: I intentionally kept all these commits to show the journey I went on but squashing this out should be straightforward.

henare:
OK I've now dramatically simplified that history into the main functional parts. If that's still too big to review I'm happy to give what @tmtmtmtm suggested in Slack a go:

>...so I'd suggest (a) pulling some of this out as distinct PRs...

views/house_wikidata.erb
```diff
@@ -18,3 +19,26 @@
...
+        sparqlQuery = "SELECT ?seat_count\n" +
+                      "WHERE { wd:" + wikidata_id + " wdt:P1342 ?seat_count }",
+        settings = {
+          headers: { Accept: 'application/sparql-results+json' },
+          data: { query: sparqlQuery }
+        };
+
+    $.ajax(endpointUrl, settings).done(function(data) {
+      if (data.results.bindings[0]) {
+        var seat_count = data.results.bindings[0][data.head.vars[0]].value;
```
tmtmtmtm: I'm a little worried about silently only taking the first value if there's more than one response. For current, national-level legislatures, the wdt: _should_ be enough to always get the up-to-date value, but I'm not sure that's always going to be true, especially as we include some historic legislatures where what the truthiest value is is less well-defined. 

I'm not sure what to suggest instead, but as this is for an internal-only page to start with, simply displaying _all_ the returned values might be enough to start with, and will let us get an idea of when (if at all) we _do_ tend to get multiple values.

henare:
@tmtmtmtm I've added handling multiple seats in 7ba67be, let me know what you think.

I've kept my incremental commits in this case because I thought since you've reviewed it already these might be helpful to see only what's changed since your last review. If, however, you'd like me to incorporate this into a rewritten history then please let me know and I'll do that instead.

henare:
>I've kept my incremental commits in this case because I thought since you've reviewed it already these might be helpful to see only what's changed since your last review. If, however, you'd like me to incorporate this into a rewritten history then please let me know and I'll do that instead.

Our [coding standards say](https://mysociety.github.io/coding-standards.html) that we should make changes in subsequent fixup commits to make re-reviews easier. So I've changed to doing that instead and once we're happy I'll do a rebase that squashes those commits.

I also added another fixup in c3bb767 to how capybara webkit's configured. I think that change make sense because it's less maintenance and we don't have a current need to restrict what URLs the tests are accessing.

henare:
This is now ready for review again :pray:

@mhl has said he's hoping to get some time to look at it soon :star2: 

I'd like to get this merged as soon as possible so we can go back to much smaller pull requests and, "_get the warm fuzzy glow of getting something into production every day :slightly_smiling_face:_", to quote @tmtmtmtm :smiley:

views/house_wikidata.erb
```diff
@@ -18,3 +19,34 @@
...
+        endpointUrl = 'https://query.wikidata.org/sparql',
+        sparqlQuery = "SELECT ?seat_count\n" +
+                      "WHERE { wd:" + wikidata_id + " wdt:P1342 ?seat_count }",
+        settings = {
+          headers: { Accept: 'application/sparql-results+json' },
+          data: { query: sparqlQuery }
+        };
+
+    $.ajax(endpointUrl, settings).done(function(data) {
+      if (data.results.bindings.length == 1) {
```
mhl: I know there are different views on this, but I prefer the Crockford style of Javascript where you explicitly hoist the variable declarations to the start of the function, since that's what'll happen anyway!

henare: Thanks @mhl, I didn't know about this!

So does that mean the function becomes this?

```javascript
    $.ajax(endpointUrl, settings).done(function(data) {
      var html, seat_count, seat_counts;

      if (data.results.bindings.length == 1) {
        seat_count = data.results.bindings[0][data.head.vars[0]].value;
        html = '<span class="seatcount">' + seat_count + '</span> seats';
      } else if (data.results.bindings.length > 1) {
        seat_counts = data.results.bindings.map(function(b) {
          return b[data.head.vars[0]].value;
        });
        html = data.results.bindings.length +
                   ' seat count values in Wikidata: <span class="seatcount">' +
                   seat_counts.join('</span>, <span class="seatcount">') +
                   '</span>';
      } else {
        html = 'Number of seats <span class="seatcount">unknown</span> in Wikidata item';
      };

      $('p.seatcount').html(html);
    });
```

```diff
diff --git a/views/house_wikidata.erb b/views/house_wikidata.erb
index e5337946..6460eeab 100644
--- a/views/house_wikidata.erb
+++ b/views/house_wikidata.erb
@@ -32,20 +32,23 @@
         };
 
     $.ajax(endpointUrl, settings).done(function(data) {
+      var html, seat_count, seat_counts;
+
       if (data.results.bindings.length == 1) {
-        var seat_count = data.results.bindings[0][data.head.vars[0]].value;
-        var html = '<span class="seatcount">' + seat_count + '</span> seats';
+        seat_count = data.results.bindings[0][data.head.vars[0]].value;
+        html = '<span class="seatcount">' + seat_count + '</span> seats';
       } else if (data.results.bindings.length > 1) {
-        var seat_counts = data.results.bindings.map(function(b) {
+        seat_counts = data.results.bindings.map(function(b) {
           return b[data.head.vars[0]].value;
         });
-        var html = data.results.bindings.length +
+        html = data.results.bindings.length +
                    ' seat count values in Wikidata: <span class="seatcount">' +
                    seat_counts.join('</span>, <span class="seatcount">') +
                    '</span>';
       } else {
-        var html = 'Number of seats <span class="seatcount">unknown</span> in Wikidata item';
+        html = 'Number of seats <span class="seatcount">unknown</span> in Wikidata item';
       };
+
       $('p.seatcount').html(html);
     });
   });
```

If :+1: I'll make the change and merge.

mhl: Yes, that's right :+1:

henare:
Made the change suggested by @mhl so merging :white_check_mark:

Issue #15639, opened 2018-01-10
Generate static site pages after JavaScript has run 


henare:
# Problem

We currently have [one page that uses JavaScript](http://everypolitician.org/needed) to progressively enhance the page by displaying up-to-date data from a remote API. It first gets the data from the GitHub API using Ruby, which is what the static site then shows on the published site. When a user visits the page the JS will also run and fetch the most up to date data from the API.

A problem with this is that the logic to get this data exists in Ruby and JS. If our static site generator ran after JS was executed we'd be able to remove the Ruby code that fetches this data, reducing the amount of code we need to maintain. Because the static site had been generated with the data included, the user experience would be the same as it currently is.

This duplication isn't a problem for one page but as part of the Democratic Commons project we're about to start [using JS to pull in a lot more data from Wikidata](https://github.com/everypolitician/democratic-commons-tasks/issues/36#issuecomment-351518555) so it makes this change a worthwhile thing to do.

# Proposed Solution

From https://github.com/everypolitician/democratic-commons-tasks/issues/36#issuecomment-351518555
>...switch to crawling these pages with Capybara / PhantomJS...

# Acceptance Criteria

The static Countries Needed page is generated populated with data, but without using the Ruby code that currently populates data for it from the GitHub API.

# Related Issues

* https://github.com/everypolitician/democratic-commons-tasks/issues/36#issuecomment-351518555
* #15637

henare:
There are two parts to the static site generation - deciding which pages to scrape, and scraping the actual pages. Both of these tasks are currently handed by `wget`. It scrapes the pages, and spiders the whole site using its recursive option (from the [seed start page](http://everypolitician.org/status/all_countries.html) we give it).

Let's discuss both separately to decide our approach.

## Deciding which pages to scrape

We currently set `wget`'s mirror option (which sets infinite recursion) and point it at the [seed start page](http://everypolitician.org/status/all_countries.html). From there it spiders pretty much every page on the site (except some pages we deliberately don't point to from the seed start page).

For this issue we could just run it on the subset of pages that need JavaScript run (currently only one page but soon to be increasing to a couple of hundred, but still by no means the whole site), keeping `wget` for the rest of the site, or we could wholesale switch to the new scraping method for every page.

Using one generation method for the whole site should make things less complex, more maintainable, and less error-prone. It's probably where we want to go eventually. However we should iterate there rather than doing it in one big go, as that would make the switch more risky and more work up-front.

[...]

PR #15640, opened 2018-01-17
Generate some static site pages after JavaScript has run using Capybara

henare:
# What does this do?

Extends the static site build process to allow some pages to be generated after JavaScript (specifically jQuery AJAX) has finished running.

# Why was this needed?

As part of https://github.com/everypolitician/democratic-commons-tasks/issues/36#issuecomment-351518555 we're about to add data to pages exclusively via JavaScript. This would result in pages missing that data using our current build process, resulting in broken pages (at least for browsers without JS).

This pull request allows us to specify pages that we want generated after JS has run and therefore the data on the page has been populated.

# Relevant Issue(s)

Closes https://github.com/everypolitician/viewer-sinatra/issues/15639.

# Implementation notes

# Notes to Reviewer

If this is working correctly then the output of the build process _should be the same_ before and after this PR. This is because the only page it's scraping has duplicated logic in the Ruby code for populating data. Once this PR is merged we can remove that duplicated logic and just build that page in JS (along with adding our new pages, when they're ready to be added to the site).

# Notes to Merger


scripts/save_javascript_pages.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+  end
+
+  def finished_all_ajax_requests?
+    page.evaluate_script('jQuery.active').zero?
+  end
+end
+
+page = PageAfterAJAX.new
+javascript_pages_to_scrape = open(ARGV[0]).read.split
```
tmtmtmtm: Probably nicer to explicitly check that there _is_ an argument, and blow up early if not.

henare:
Thanks @tmtmtmtm for the detailed and very useful feedback :star2:

I've responded below to the detail but here's a summary of the actions I'd like to take, in roughly the order they should be done in:

* Point this PR at `master` and remove the unnecessary dependency on https://github.com/everypolitician/viewer-sinatra/pull/15637
* Refactor class to use Pathname
* Refactor class to be less imperative
* Move the `save_javascript_pages.rb` script into the application somewhere, renaming the class in the process
* Remove the unnecessary `javascript_pages_to_scrape.txt` endpoint
* Have the process run from a Rake task

As this is a pretty significant rework I don't think there's going to be a lot of benefit in me using fixup commits so I'll probably rewrite history a bit as part of this process.

## Approach

>I'm not quite following the approach of serving up `javascript_pages_to_scrape.txt` through the app, rather than simply having that be a text file entirely outside the app that the script reads directly.

We want a way for the app to generate a list of URLs (e.g. a set of pages for every country we have) so I think I was thinking, "how can we communicate this to a process [the script] that's running outside the application?". In my initial notes for this work I was calling this a Rake task but then I don't even mention that in my [discussion on how to approach this part](https://github.com/everypolitician/viewer-sinatra/issues/15639#issuecomment-356815495). I think the obvious thing is to bring this process into the application itself, as I'd been originally thinking with a Raketask.

>If this is meant to be a parallel with how the `wget` equivalent works, I fear that that's a slightly misleading consistency that doesn't buy us much, and makes these seem more similar than they actually are. The `wget` base is a page because that version is a recursive process of fetching (and saving) a given page and all other pages linked from that, whereas here you're simply working down a single list non-recursively.

:+1:

## Implementation

>The implementation of `PageAfterAJAX` is also little odd to me (or possibly the naming, depending on what the class is actually supposed to represent). The naming implies that it's a Page, but it actually seems to be a wrapper around Capybara which you then feed pages to.

It's probably the implementation _and_ the naming :smile: It wasn't supposed to be a Page like in the application so the naming is definitely where the problems start.

If we bring it into the application as I suggest above then I'll try come up with a better name, and one that doesn't imply it's a Page.

>In general a constructor that doesn't take any arguments is a bit of an anti-pattern, especially when lots of the methods then take a consistent argument (in this case `url`). That implies that either this class _should_ actually represent a Page, or at least that there's a `Page` class lurking in here, looking to escape. (I don't know whether creating a new Capybara::Session is expensive enough to want to share that across the URLs, or whether it's lightweight enough to simply create a new one each time)

Yes, I agree this looks weird. It _is_ because the intent is to share a Capybara::Session across URLs because these are expensive (akin to booting a browser AIUI).

>The implementation is also quite imperative at the minute. That's not necessarily a bad thing, and we certainly shouldn't try to needlessly shoe-horn it into an OO pattern, but giving it the trappings of being in a Class, but then not really being one, is a little confusing. So, for example, rather than `create_parent_directories`, you could have a `containing_directory` method which also takes care of making sure that the directory exists before returning it. (That's strictly a violation of command-query separation, but I think it'd be OK here.)

Good call. I don't natively think in OO so it's not surprising it's quite imperative. I'll make that change and try to keep going with a refactor in that direction.

>You should also be able to simplify some of the implementation by using [Pathname](https://ruby-doc.org/stdlib-2.3.0/libdoc/pathname/rdoc/Pathname.html) everywhere, rather than doing so much string manipulation. It has built-ins for getting parts out of the path, transforming the name, creating the directory, writing the file, etc.

Great! I've not used that library so I'll give it a shot.

## Setup 

>Having this branch off #15637 seems like a slightly needless dependency here (or a dependency in the wrong order). Is there a reason this PR can't go live before that one? It seems that whichever commit(s) are needed for both could be cherry-picked into here, and go live with whichever is merged first, if introducing them in their own PR first doesn't make sense.

Thank you for picking up on this. I don't know what my thinking was here. I'm guessing it was just because these bits of work had become conflated in my head, but of course they're not in code. I'll fix that up too.

henare:
I've done a pretty major reworking of this now. It's called from a Rake task and the class now just takes an array of URLs to build. I think this all makes it a lot easier to understand and simplifies the implementation.

This still allows us to easily expand the set of pages we want to build by inserting more URLs into the array we pass to StaticSiteGenerator.

@tmtmtmtm could you please review this again? :sunflower:

lib/static_site_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+require 'pathname'
+require 'uri'
+
+class StaticSiteGenerator
+  attr_reader :page, :urls
+
+  def initialize(urls:)
+    Capybara::Webkit.configure(&:allow_unknown_urls)
+    @page = Capybara::Session.new(:webkit)
+    @urls = urls.map { |u| URI.parse(u) }
```
tmtmtmtm: NB: our practice in this repo is to avoid doing work directly in the constructor — most of these seem like things that can be deferred to the methods that ask for them.

henare:
Sorry this has been such a pain to review @tmtmtmtm :hurtrealbad: 

As suggested I went to add tests for that class. Straight away I came up against a road block in that most of that class's methods were private and so difficult to test.

Instead of battling through that I tried to take what you said in the rest of your comment and rethink each part in terms of what it is. What I've come up with is 358001c87369f51150821a721b6da556ed0a585c which I think is a bit simpler.

As I mention in the commit it feels like some of the complexity has just been pushed in to the rake task so there's probably room for improvement there. Do you have any suggestions for things I could try?

We also a need to test the new Browser class. Since the methods on that class don't _do_ a whole lot by themselves I think this will mean some kind of integration-level test with a page fixture.

henare:
Closing ancient pull requests :moyai:

PR #15642, opened 2018-01-19
Fix test deprecation warnings about not using assert_nil

henare:
# What does this do?

Some tests were causing this deprecation warning:

>DEPRECATED: Use assert_nil if expecting nil from $FILENAME. This will fail in Minitest 6.

This PR fixes those deprecation warnings by updating those tests to use `#must_be_nil` instead of `#must_equal nil`.

# Why was this needed?

These warnings were added in [minitest 5.10.0](https://github.com/seattlerb/minitest/blob/master/History.rdoc#5100--2016-11-30) and we [upgraded to minitest > 5.10.0 recently](https://github.com/everypolitician/viewer-sinatra/pull/15633/files#diff-e79a60dc6b85309ae70a6ea8261eaf95L51).

Deprecation warnings on master: https://travis-ci.org/everypolitician/viewer-sinatra/builds/316216105#L896
Now no deprecation warnings: https://travis-ci.org/everypolitician/viewer-sinatra/builds/330669752#L1041

# Relevant Issue(s)

# Implementation notes

# Screenshots

# Notes to Reviewer

# Notes to Merger

### DIFF
--- a/t/page/term_table.rb
+++ b/t/page/term_table.rb
@@ -91,7 +91,7 @@ describe 'TermTable' do
       it 'has a single legislative membership' do
         af.legislative_memberships.count.must_equal 1
         af.legislative_memberships.first.start_date.must_equal '2013-10-29'
-        af.legislative_memberships.first.end_date.must_equal nil
+        af.legislative_memberships.first.end_date.must_be_nil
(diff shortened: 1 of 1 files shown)


henare:
Thanks for the merge @tmtmtmtm :star2:

Issue #15643, opened 2018-01-23
[Wikidata] Create countries list page


henare:
# Problem

As part of https://github.com/everypolitician/democratic-commons-tasks/issues/36 we want to display a list of countries from Wikidata.

# Proposed Solution

Recreate the home page from https://github.com/tmtmtmtm/wikidata-legislature-browser/ albeit using client-side JavaScript. Have this page captured by our JS-enabled process (pending merge in #15640).

Put this list at `/wikidata`.

# Acceptance Criteria

# Not Required

# Prerequisites

# Related Issues

* https://github.com/everypolitician/democratic-commons-tasks/issues/36
* #15640

# Due Date

# Design?

# Text?

# Bloggable?

henare:
At the start of February we decided to switch the focus of our project to developing this in https://github.com/everypolitician/legislative-explorer/ so closing this.

PR #15646, opened 2018-01-30
[WIP] Wikidata countries list page

henare:
# What does this do?

Creates a new page at `/wikidata/countries.html` which displays the same thing as the wikidata-legislature-browser. The page is populated using JavaScript on the client instead of the Ruby/Sinatra backend.

# Why was this needed?

We're recreating the pages that the wikidata-legislature-browser prototype has but implementing them in JavaScript on the client so our static pages can have more up to date data from the Wikidata API. This is the first of those pages.

# Relevant Issue(s)

#15643

# Implementation notes

This uses a [JavaScript SPARQL API client](https://github.com/wikimedia/wikidata-query-gui/blob/cfb3a0d36f12cf72ceb7d8c4887ad49ef0a8f5b1/wikibase/queryService/api/Sparql.js) from the wikidata-query-gui project. It makes doing SPARQL Wikidata queries from JavaScript easier. It's GPLv2 licensed so I think we're fine to use it in this project.

# Screenshots

Ruby prototype left, this PR right:

screenshot_from_2018-01-23_16-47-28

# Notes to Reviewer

This PR doesn't currently have any tests. Are integration-level tests from minitest fine or do we want to set up some JavaScript testing?

# Notes to Merger

Because this page isn't currently set up to be scraped onto the static site this won't be visible on the deployed site.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -99,6 +99,11 @@ get '/needed.html' do
   erb :needed
 end
 
+get '/wikidata/countries.html' do
+  @page = Page::WikidataCountries.new
+  erb :"wikidata/countries"
+end
+
 get '/*.css' do |filename|
   scss :"sass/#{filename}"
 end
--- a/lib/page/wikidata_countries.rb
+++ b/lib/page/wikidata_countries.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Page
+  class WikidataCountries
+    def title
+      'EveryPolitician'
+    end
+  end
+end
--- a/public/javascript/vendor/Sparql.js
+++ b/public/javascript/vendor/Sparql.js
@@ -1,0 +1,488 @@
+var wikibase = wikibase || {};
+wikibase.queryService = wikibase.queryService || {};
+wikibase.queryService.api = wikibase.queryService.api || {};
+
+wikibase.queryService.api.Sparql = ( function( $ ) {
+	'use strict';
+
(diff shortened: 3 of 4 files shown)


mhl:
Hi @henare - on the Javascript testing question, I do think it would be good idea to add a couple of tests, (e.g. perhaps one just testing `entityID` as an example of library code and one testing `updateTitles` an example of testing code affecting the DOM) since then there'll be so much less friction for adding more tests in the future.

I'm not at all experienced when it comes to Javascript testing, but I did find it relatively easy previously to set up a combination of PhantomJS, Mocha and Chai for another project - the pull request to do that is here: https://github.com/everypolitician/membership_matrix/pull/15 (n.b. I don't think that includes any tests that had to make assertions about changes to the DOM, which might be more involved...) I think I chose that combination largely on the basis of what @chrismytton had used for PopIt, so he might have some thoughts too, but I'd probably start with something similar.

(In case it's of interest, I do have a nice ebook about Javascript testing, which is: https://jstesting.jcoglan.com/ (it uses jstest, which was written by the author, for its examples, but I think it's generally interesting no matter what libraries you're using) - also as bonus, appendix A has one of the best explanations of the different ways of invoking Javascript functions that I've seen.)

henare:
Since we've decided this work should continue in https://github.com/everypolitician/legislative-explorer/ I'll close this PR.

PR #15647, opened 2018-02-02
Install `tidy` from package on Travis

henare:
# What does this do?

Previously we had a script that compiled a recent HTML Tidy from source
as the version that ships with Ubuntu Trusty is ancient. This switches
us to using the version that ships in the Debian sid repo which is the
same version we were compiling from source:

https://packages.debian.org/sid/tidy

# Why was this needed?

Removing the custom script and not compiling something from source makes
things a little simpler for us and as a side effect should speed up the
build slightly.

# Relevant Issue(s)

# Implementation notes

# Screenshots

# Notes to Reviewer

# Notes to Merger


tmtmtmtm:
Thanks @henare — once we start getting into this sort of systems-wrangling it's straying outside my field though. Perhaps @chrismytton (who added this approach in the first place) might be able to give it a quick thumbs-up if it seems sensible?

henare:
Thanks for checking @tmtmtmtm and thanks for reviewing @chrismytton :shipit:

PR #15651, opened 2018-04-03
Vagrant ruby capybara updates

zarino:
# What does this do?

Enables you to run `vagrant up` in the project directory, and be given a working vagrant VM, so you can `vagrant ssh` in and run `foreman start` to work on the EveryPolitician front-end locally.

# Why was this needed?

The Vagrant provisioner was out of date from #15649  – it requested a version of Ruby that caused an error from Bundler, and it lacked a bunch of native packages that `capybara-webkit` requires.

# Relevant Issue(s)

# Implementation notes

Tested with Vagrant 2.0.2 and Virtualbox 5.2.8r121009 on macOS 10.13.3.

# Screenshots

# Notes to Reviewer

# Notes to Merger

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,11 @@ Finally, follow the instructions displayed by the virtual machine.
 
 ### Run without Vagrant
 
-You can also run this project locally without using Vagrant. First [install Foreman globally](https://github.com/ddollar/foreman#installation):
+You can also run this project locally without using Vagrant, but you will need
+to make sure the "required packages" from `script/provision.sh` are available
+locally, before you begin.
+
+Then [install Foreman globally](https://github.com/ddollar/foreman#installation):
 
 ```bash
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -50,7 +50,11 @@ Finally, follow the instructions displayed by the virtual machine.
 
 ### Run without Vagrant
 
-You can also run this project locally without using Vagrant. First [install Foreman globally](https://github.com/ddollar/foreman#installation):
+You can also run this project locally without using Vagrant, but you will need
+to make sure the "required packages" from `script/provision.sh` are available
+locally, before you begin.
+
```
tmtmtmtm: :+1: for this addition. It would probably be useful to also add what these are likely to be on something other than debian (e.g. brew), but that can be for later!

PR #15652, opened 2018-04-03
Add democratic commons links

wrightmartin:
# What does this do?
Adds information to the EveryPolitician website about the Democratic Commons project it is now part of. It's barebones and I expect when assets exist we will come back and add a logo or graphic treatment.

# Why was this needed?
Previously there were no links at all

# Relevant Issue(s)

# Implementation notes
None, just basic HTML and CSS changes

# Screenshots
## Homepage
screen shot 2018-04-03 at 15 51 37

## Country page
screen shot 2018-04-03 at 15 51 30

# Notes to Reviewer

# Notes to Merger

### DIFF
--- a/t/web/house.rb
+++ b/t/web/house.rb
@@ -21,6 +21,15 @@ describe 'House Page' do
     end
   end
 
+  describe 'when displaying democratic commons footer' do
+    let(:link) { subject.css('.demo-commons-promo') }
+    before     { get '/australia/representatives/' }
+
+    it 'should link to mySociety Democratic Commons page' do
+      link.css('a').last[:href].must_equal 'https://www.mysociety.org/democracy/democratic-commons/'
+    end
+  end
+
   describe 'HTML validation' do
     it 'has no errors in the house page' do
       last_response_must_be_valid


views/homepage.erb
```diff
@@ -14,17 +14,22 @@
...
             <h3>What is EveryPolitician?</h3>
             <p>Data about every national legislature in the world, freely available for you to use:</p>
-            <a href="http://docs.everypolitician.org/" class="button button--small">Read more…</a>
+            <a href="http://docs.everypolitician.org/">EveryPolitician documentation</a>
         </div>
         <div class="column-one-third">
             <h3>Contribute data</h3>
             <p>Missing data for your country? Here's how to get that changed:</p>
-            <a href="http://docs.everypolitician.org/contribute.html" class="button button--small button--secondary">Read more…</a>
+            <a href="http://docs.everypolitician.org/contribute.html">How to contribue</a>
```
tmtmtmtm: contribu*t*e

wrightmartin:
Thanks for the feedback @tmtmtmtm - typo fixed, fixup gone. 

I've not written a test before but I took a run at it - see what you think

wrightmartin:
> More fundamentally, though, I'm not convinced that this change makes sense in isolation. It gives a quite confusing message about the project, particularly as it's placed right next to a "How to contribute" link that mentions nothing at all about Wikidata or the Commons, and indeed gives conflicting information.

Make sense to me - but in this case you're talking to the monkey and not the organ grinder - Rich will be the one to talk to regarding that

wrightmartin:
oh, I see the test is failing, although they passed locally - we are now outside of my area of competence

wrightmartin:
> This is because one of the tests is to essentially try to generate the static site and make sure we're able to do so, but you've removed this image, even though it's still linked from views/house.erb.

Aha, hadn't spotted that, have replaced it with the same as the country template

tmtmtmtm:
Great! Looks like the only thing travis is complaining about now is the awkward spacing in `before      { get '/iran/' }` (NB: running the default rake task — i.e. `bundle exec rake` — will run not only the tests, but also the rubocop / reek checks etc, which will show you these)

rich-cassidy:
@tmtmtmtm @wrightmartin 

> More fundamentally, though, I'm not convinced that this change makes sense in isolation. It gives a quite confusing message about the project, particularly as it's placed right next to a "How to contribute" link that mentions nothing at all about Wikidata or the Commons, and indeed gives conflicting information.

With regards to this, what accompanying change would be helpful to eliminate / reduce the confusion? It is changing the 'Contribute data' section to mention add info into Wikidata and/or pointing to an appropriate place on Wikidata (perhaps the Wikiproject page)?

tmtmtmtm:
@rich-cassidy https://github.com/everypolitician/everypolitician-docs/issues/43 is tracking the related changes that we would need to make to the docs

tmtmtmtm:
> All changes made and tests added

Thanks @wrightmartin — I suspect we might need to tweak some of this again before putting it live once the wider issue is resolved, but maybe by then we'll also have accompanying assets too ;)

PR #15659, opened 2018-11-30
Signpost Wikidata on the homepage

tmtmtmtm:
We're rewriting all the docs to make it much clearer that the goal is
for the vast majority of this information to live in Wikidata. That
rewrite is still in progress, but keeping the home page as it was is out
of date with the rest of our messaging around this, and just as
confusing as having it link to out of date docs.

https://github.com/everypolitician/everypolitician-docs/issues/43


tmtmtmtm:
Fixing the unrelated test failures is happening in https://github.com/everypolitician/viewer-sinatra/pull/15660

PR #15673, opened 2019-03-11
Remove needless space from README

tmtmtmtm:
Remove a spurious space that has crept in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ bundle exec rake
 
 The project uses [Sinatra](http://www.sinatrarb.com), an ultra-minimal Ruby MVC web app framework.
 
-In development, changes to `app.rb` and Ruby files in the `lib` directory are automatically reloaded by  [`Sinatra::Reloader`](http://sinatrarb.com/contrib/reloader.html). You do not need to restart the webserver process.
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
@sagepe Looks like we have another case of travis-rot :(

The problem here seems to be with `scripts/release.sh`. As EveryPolitician.org is a static site, we try a full-rebuild of the site with every PR to make sure there are no problems, but travis looks like it's encountering a slightly odd problem when it tries doing that (the key search term in the output is "Rack app error handling request") where it looks like it's getting truncated popolo JSON. This is consistently failing across re-runs, and different PRs, though in some runs it fails at Thailand; in others Thailand is fine, and then it stumbles at the Vatican (e.g. in https://travis-ci.org/everypolitician/viewer-sinatra/builds/504860625).

The script runs with no issues on my laptop, so I can't even drop into a `pry` session and go spelunking. Any ideas?

tmtmtmtm:
Hopefully rebasing this over the new master (post-#15675) will do the trick…

tmtmtmtm:
Woohoo! I don't really understand *why* the `sudo: false` would cause the truncated-JSON problems we've been dealing with here, but yay for @chrismytton for spotting it!

PR #15675, opened 2019-05-09
Make travis happy again

tmtmtmtm:
Travis has been broken for quite some time now, and it's not clear what exactly is going awry.

@chrismytton noticed that `sudo: false` is now a deprecated configuration option, and should be removed:
  https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments

We *might* need (or want) to also explicitly specify a `dist` but initially we can try simply using the default to see how that goes.


tmtmtmtm:
@sagepe @chrismytton Simply removing the `sudo: false` is looking promising. Travis is still failing, but this now seems to solely be due to the CVE against the version of `nokogiri` we're using. Under normal circumstances we'd want to fix *that* issue in a separate PR, but we've got a combined dependency here now, so hopefully also including a `bundle update` here will be enough to get everything back on track…

tmtmtmtm:
Grr. A *full* `bundle update` causes _other_ problems (mainly because that gets us an updated `rubocop` with some new rules we're not currently in compliance with), so I've backed that out and will restrict this PR to only security-related updates. If this gets us to a clean setup again, we can then look at updating other gems separately, etc.

tmtmtmtm:
🤸‍♀ 👯‍♂ 🚀

chrismytton:
Hooray!

PR #15683, opened 2020-03-17
Remove cookie storage from GA

crowbot:
# What does this do?

Remove cookie tracking from GA

# Why was this needed?

We don't want to track individual users

# Relevant Issue(s)

https://github.com/mysociety/orgsites/issues/1124

# Notes to Merger

Could you merge and deploy this if it looks ok and it's reasonably easy, and let me know if it's not easy? Thanks!


chrismytton:
This is now live: https://everypolitician.org/

